PR #3, opened 2019-09-18
Initial Boilerplate

paulobmarcos:
This is a stripped down version of one of our projects.

I think that it currently represents what we achieved already, but we sure can upgrade it to always be better and more complete!

One thing that I would like you to focus is that in this boilerplate I provided a "few" examples to better explain the structure we usually follow.

We should also review the README. This one was a mix between `react-with-moxy` and the READMEs we have been using recently.

Let's get it started 💪🎉🥳


babel.config.js
```diff
@@ -0,0 +1,16 @@
...
+    api.cache(false);
+
+    return require('babel-preset-moxy/end-project')(api, {
+        modules: process.env.NODE_ENV === 'test' ? 'commonjs' : false,
+        react: true,
+        loose: false,
+        targets: {
+            // Maintain the node version as same as the one in Dockerfile
+            node: '12.6.0',
+            browsers: ['extends browserslist-config-google'],
```
ivolimasilva: I think we should have this configuration point to the browsers' target as the **PostCSS** configuration. One option is to pass this to the `package.json` and have this file require the `package.json` and pass
```
browsers: require('./package.json').browserslist,
```

next.config.js
```diff
@@ -0,0 +1,164 @@
...
+                    name: dev ? 'playback/[name].[ext]' : 'playback/[name].[hash:15].[ext]',
+                    publicPath: `${assetPrefix}/_next/static/chunks/media`,
+                    outputPath: 'static/chunks/media',
+                    emitFile: !isServer,
+                },
+            });
+
+            // SVG files
+            config.module.rules.push({
+                test: /images\/.*.svg$/,
```
ivolimasilva: In here we're forcing the users to use the folder `images/` to put SVGs that aren't considered to be inline. In conversation with @satazor we referenced that we shouldn't use folder specific rules but file specific rules (since we're dealing basically with file extensions).
With that said, and with the developments made in [@moxy/next-common-files](https://github.com/moxystudio/next-common-files), I think it's better to have something like `.data-url.svg` for "normal" SVGs and use `.svg` for inline SVGs. Or another name.

now.json
```diff
@@ -0,0 +1,6 @@
+{
```
ivolimasilva: This file is specific for [now.sh](https://zeit.co/) deploys. I know there's a description on the **README** about this I don't think this should come in the boilerplate.

Also, using version **2** means that we loose SSR since the app will be served as serverless.

package.json
```diff
@@ -0,0 +1,98 @@
...
+    "postcss-preset-moxy": "^4.3.0",
+    "raw-loader": "^3.0.0",
+    "rimraf": "^2.6.3",
+    "standard-version": "^6.0.1",
+    "stylelint": "^10.1.0",
+    "stylelint-config-moxy": "^6.2.0",
+    "svg-css-modules-loader": "^1.6.1",
+    "svgo-loader": "^2.2.1"
+  },
+  "husky": {
```
ivolimasilva: Bringing the discussion up about having these configurations here instead of specific files.

I, for one, prefer having it in specific files since I believe it helps finding the information more quickly and it doesn't pollute the `package.json` with information about tooling.

Paging @satazor for his input.

package.json
```diff
@@ -0,0 +1,98 @@
...
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+    }
+  },
+  "standard-version": {
+    "scripts": {
+      "posttag": "git push --follow-tags origin master"
```
ivolimasilva: I believe it's more obvious to have a _npm script_ (such as `npm run release`) to run `standard-version` as it seems the recommended way by [standard-version](https://github.com/conventional-changelog/standard-version#as-npm-run-script) and have a `postrelease` _script_ to push the tag, removing the need for this object to exist.

postcss.config.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable prefer-import/prefer-import-over-require */
```
ivolimasilva: As said in the comment of `babel.config.js`, I think this configuration should look to the same target as that one.

Also, we should inlining any fonts used to avoid flashes of layout.

server.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable prefer-import/prefer-import-over-require */
+
+const path = require('path');
+const express = require('express');
+const next = require('next');
+const compression = require('compression');
+const expressStaticGzip = require('express-static-gzip');
+
+const host = process.env.HOST || '127.0.0.1';
```
ivolimasilva: Having this will keep us from accessing our IP in the same network which is very useful when testing in other devices.

src/app/App.js
```diff
@@ -0,0 +1,72 @@
...
+const SEO_DATA = {
+    url: 'https://{project-domain}',
+    title: '{project-name}',
+    description: '{project-description}',
+    keywords: ['{project-keyword-1}', '{project-keyword-2}'],
+    image: { src: seoImage, width: 1200, height: 630 },
+};
+
+class App extends BaseApp {
+    static async getInitialProps({ Component, ctx }) {
```
ivolimasilva: We can remove this method and only add when needed: [Opt-out of Automatic Prerendering](https://github.com/zeit/next.js/blob/master/errors/opt-out-automatic-prerendering.md)

src/shared/components/contacts/Contacts.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { contacts } from '../../data';
+
+const Contacts = () => (
+    <>
```
ivolimasilva: It's preferred to use `<>`? I think it's quite weird and it's inconsistent when you need to use a `key` prop.

I think we should also add a file of conventions (or only maybe internally) about stuff like this.

src/shared/data/contacts.js
```diff
@@ -0,0 +1,6 @@
+const contacts = {
```
ivolimasilva: I believe it's better to have "data" files next to the code that uses it. In previous projects we used `.data.js` next to the code and ignore it for tests/coverage.

This way, it's easier to find static data about the code you're working on. Also, using a `data` folder means that if a project needs a lot of static data files, this folder will have a lot of files which hardens when searching for one specific.

src/shared/styles/base/layout.css
```diff
@@ -0,0 +1,47 @@
...
+
+/* ==========================================================================
+   Proper box-sizing model
+   ========================================================================== */
+
+*,
+*::before,
+*::after {
+    box-sizing: border-box;
+    cursor: none;
```
ivolimasilva: Errr, I think this shouldn't be here 😜

src/shared/styles/custom-medias.css
```diff
@@ -0,0 +1,14 @@
+/* ==========================================================================
+   Custom media queries
+
+   Use it like so: @media (--layout-small) { ... }
+   ========================================================================== */
+
+/* Use these if you developing desktop first and want to target lower resolutions */
+
+@custom-media --layout-lte-small-mobile (width <= 20em); /* 320px * 0.0625 */
```
ivolimasilva: I believe `375px` is a better breakpoint to target smaller devices and not the "normal" ones, like iPhone 6S.

README.md
```diff
@@ -1,2 +1,191 @@
...
+
+## Deployment
+
+### Staging
+
+There's a CI/CD pipeline configured for staging, allowing you to quickly preview outcome of changes.
+
+If the build succeeds, any commit will be deployed on [now](https://zeit.co/now) and will have a unique URL for you to preview the deployment. Moreover, branches have fixed URLs, meaning they will stay the same even if they change over time in terms of commits.
+
+These URLs will are available directly on GitLab, thanks to `now`'s tight integration.
```
ivolimasilva: I think we shouldn't advice this integration since it requires the connection between the project and a team member. The integration doesn't work with `now` teams so we've been using a CI configuration to deploy with a generic token and not in the name of a team member.

package.json
satazor: I don’t have a strong opinion on this really. Although, if I may give one, I would prefer to do it just one way, that is, either all in package.json or all in separate files.

package.json
satazor: Yes we should have a release npm script that runs standard-version. I’m also in favor of having a postrelease script instead of having a separate key in here.

src/shared/components/contacts/Contacts.js
satazor: I prefer `<>` because we avoid confusing Fragment with a react component, which it isn’t. Keys in this context are very rare, and for those cases we may still use Fragment.

src/shared/data/contacts.js
satazor: Agree

README.md
satazor: If we remove now.json, we should remove this section altogether. Perhaps setting up now should be a recipe.

babel.config.js
satazor: One thing to consider is to change both eslint and stylelint presets to respect a brownish file or prop in package json.

Anyway, I agree to centralize this definition.

server.js
paulobmarcos: Maybe we can change the default to `0.0.0.0`?

src/shared/components/contacts/Contacts.js
paulobmarcos: I also thought it was strange at first, but the ease of use got me and after some time it didn't bother me as much. I also agree with @satazor, I think keys will be an edge-case and maybe we can re-think the logic to embrace this better or, in the last case, use Fragment.

src/shared/data/contacts.js
paulobmarcos: I have a different opinion on that. Imagine that for some reason someone not really into the project wants to change some static data, like changing a date, an email or a phone number. I think that a file with data is easier to understand and manage, without needing to know our file structure and figuring where the information actually is.

src/shared/styles/custom-medias.css
paulobmarcos: I believe that `320px` was the "problem" in most projects we work in. So I think that this should be a breakpoint to target small phones like `iPhone SE`. `375px` is the viewport width of an `iPhone X` and I think that doesn't represent "small" devices.

README.md
paulobmarcos: I agree, this should be a recipe. Where do you think we should put recipes?
Should this repo have a folder like `boilerplate` or `next-with-moxy` where it has all necessary files and a folder name `recipes` where it has a README pointing to specific recipe READMEs?

package.json
paulobmarcos: I think we can try having it on specific files. It might not clutter the `package.json` as much.

src/app/App.js
paulobmarcos: Maybe we should take this to a recipe?

src/shared/components/contacts/Contacts.js
paulobmarcos: -- We should check if there is an eslint rule for this.

now.json
paulobmarcos: -- We should analyse other alternatives like Netlify, but for now we should keep this since it is currently our way of staging.

src/shared/data/contacts.js
paulobmarcos: -- open issue

src/shared/styles/custom-medias.css
paulobmarcos: -- this should also be part of the design system

babel.config.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = (api) => {
+    api.cache(false);
+
+    return require('babel-preset-moxy/end-project')(api, {
+        modules: process.env.NODE_ENV === 'test' ? 'commonjs' : false,
+        react: true,
+        loose: false,
+        targets: {
+            // Maintain the node version as same as the one in Dockerfile
+            node: '12.6.0',
```
paulobmarcos: --- node: 'current'

babel.config.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable prefer-import/prefer-import-over-require */
+
+module.exports = (api) => {
+    api.cache(false);
+
+    return require('babel-preset-moxy/end-project')(api, {
+        modules: process.env.NODE_ENV === 'test' ? 'commonjs' : false,
+        react: true,
+        loose: false,
```
paulobmarcos: -- should we make loose: false the default in our babel-preset?

README.md
```diff
@@ -1,2 +1,191 @@
 # next-with-moxy
+
 MOXY's boilerplate to create Next.js based applications
+
+This project uses [Next.js](https://nextjs.org) and was bootstrapped manually using MOXY's most used presets for tooling.
+
+## Motivation
```
paulobmarcos: Add new line here.

now.json
paulobmarcos: Opened issue. I will keep `now` for now since it is our bet at the time.

src/shared/components/contacts/Contacts.js
paulobmarcos: We have this rule we can use: `jsx-fragments`. 
Where:
image

README.md
paulobmarcos: After speaking with @satazor and @ivolimasilva, we will start by creating a `recipes` folder in this repo with all recipes. Together will also be a `boilerplate` folder.

src/app/App.js
paulobmarcos: Done, created a recipe for automatic prerendering

README.md
paulobmarcos: Done

src/shared/components/contacts/Contacts.js
paulobmarcos: Done

boilerplate/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "project-name",
+  "version": "0.1.0",
```
ivolimasilva: Start with `0.0.0`?

README.md
```diff
@@ -1,2 +1,49 @@
...
+- [Installation and setup](#installation-and-setup)
+- [Used in](#used-in)
+
+## Installation and setup
+
+Use the boilerplate available in the `boilerplate` folder since it provides everything needed to start.
+
+Before using, change some information specific to the project. In the `pacakge.json` replace `name`, `description` and `repository.url`. 
+
+Detailed information about the boilerplate is available in its [README](boilerplate/README.md).
```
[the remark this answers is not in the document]
paulobmarcos: I don't think that applies here since even if you fork it, you wold have to delete other stuff that you don't need. Having it in the boilerplate folder means that you can simply download that and thats it. I think that @satazor can provide a better insight on this.

boilerplate/now.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
paulobmarcos: We are not removing *Now* for now since this is for us, MOXY, and currently it is our only/preferred why for staging.

boilerplate/src/app/App.js
```diff
@@ -0,0 +1,62 @@
...
+                    <meta property="og:image:width" content={ SEO_DATA.image.width } />
+                    <meta property="og:image:height" content={ SEO_DATA.image.height } />
+                    {/* Twitter */}
+                    <meta property="twitter:card" content="summary_large_image" />
+                    <meta property="twitter:url" content={ SEO_DATA.url } />
+                    <meta property="twitter:title" content={ SEO_DATA.title } />
+                    <meta property="twitter:description" content={ SEO_DATA.description } />
+                    <meta property="twitter:image" content={ SEO_DATA.image.src } />
+                </Head>
+                <Component { ...pageProps } />
```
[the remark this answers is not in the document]
paulobmarcos: I don't like to force a Header and Footer. We don't know if projects will need it or not. Sometimes they are needed for some routes only. So, I would leave that for each team to decide.

boilerplate/package.json
paulobmarcos: Done :+1:

boilerplate/src/app/App.js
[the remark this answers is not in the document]
ivolimasilva: Yes, but I would be an example to guide on where to put them (here in App so it doesn't re-render in each page). However, that requires some more exampling to do and I don't think it's necessary but it would definitely be a nice to have.

The other point however is something that I would like to have your opinion on since I do think it's a good way to force a good practice.

boilerplate/src/shared/styles/mixins/typography.css
```diff
@@ -0,0 +1,110 @@
+@import "../custom-medias";
+
+/* ==========================================================================
+  Fonts
+  ========================================================================== */
+
+@define-mixin font-montserrat $fontWeight: 400 {
+    font-family: "Montserrat", sans-serif;
+    font-weight: $(fontWeight);
```
nlfonseca: you can put `font-weight: $fontWeight;`

boilerplate/src/shared/styles/mixins/typography.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/* ==========================================================================
+  All typography variants
+
+  $unit - Either em or rem
+  ========================================================================== */
+
+@define-mixin typography-numbers $unit {
```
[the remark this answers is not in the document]
nlfonseca: yes we can, maybe with quotation marks 
`@define-mixin typography-numbers $unit: "rem" {`

boilerplate/src/shared/styles/mixins/typography.css
```diff
@@ -0,0 +1,110 @@
...
+    line-height: calc(5.8em / 6);
+
+    @media (--layout-lte-tablet) {
+        font-size: 4.8$(unit);
+        line-height: calc(5.8em / 4.8);
+    }
+}
+
+@define-mixin typography-h1 $unit {
+    @mixin font-montserrat 600;
```
nlfonseca: Normally I think is better separate mixins from the rest of the rules.
What do you think?

```
@define-mixin typography-h1 $unit {
     @mixin font-montserrat 600;

     font-size: 4.2$(unit);
     line-height: calc(5.8 / 4.2);
}

File: README.md (hunk 18 above)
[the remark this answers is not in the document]
Comment (nlfonseca) [reply README.md]: I agree with @ivolimasilva. Yes, maybe we will always have to delete other stuff that we don't need, but this folder will be always deleted, so it will be faster for the user to start coding.

File: README.md [hunk 24]
```diff
@@ -1,2 +1,49 @@
...
+
+## Recipes
+
+[Recipes](recipes/README.md) provide detailed infomartion on how to use certain features that are not enabled by default.
+
+## Used In
+
+- [Platforme](http://platforme.com/)
+- [dashdash](https://dashdash.com)
```
nlfonseca: Is okay to have this public?

README.md
```diff
@@ -1,2 +1,49 @@
...
+## Table of Contents
+
+- [Installation and setup](#installation-and-setup)
+- [Used in](#used-in)
+
+## Installation and setup
+
+Use the boilerplate available in the `boilerplate` folder since it provides everything needed to start.
+
+Before using, change some information specific to the project. In the `pacakge.json` replace `name`, `description` and `repository.url`.
```
rafaelramalho19:
```suggestion
Before using, change some information specific to the project. In the `package.json` replace `name`, `description` and `repository.url`. 
```

README.md
```diff
@@ -1,2 +1,49 @@
...
+
+Use the boilerplate available in the `boilerplate` folder since it provides everything needed to start.
+
+Before using, change some information specific to the project. In the `pacakge.json` replace `name`, `description` and `repository.url`. 
+
+Detailed information about the boilerplate is available in its [README](boilerplate/README.md).
+
+## Recipes
+
+[Recipes](recipes/README.md) provide detailed infomartion on how to use certain features that are not enabled by default.
```
rafaelramalho19:
```suggestion
[Recipes](recipes/README.md) provide detailed information on how to use certain features that are not enabled by default.
```

boilerplate/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Options
+
+Every option is passed to the `jest` command, however here are the most useful ones:
+
+##### `<expression>`
+
+If passed an expression, **Jest** will compare every file with that expression and only run tests of those whose match is positive.
+
+##### `--coverage`
```
rafaelramalho19: Shouldn't we have the `--coverage` flag by default in the `test` script?

boilerplate/jest.config.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+    // Snapshots
+    setupFilesAfterEnv: ['./node_modules/jest-enzyme/lib/index.js'],
+    // Coverage
+    collectCoverageFrom: [
+        'src/**/*.js',
+    ],
+    coverageThreshold: {
+        global: {
+            branches: 80,
```
rafaelramalho19: I would advocate for 100 in the default values 😄

rafaelramalho19:
How about integration with some kind of generator system (e.g. yeoman) to create a new project with this template and replace all the variables (e.g. `package-name`)?

satazor:
@rafaelramalho19 we have discussed that, but some concerns were raised in the time it would be needed to find the right tool, write the scripts and maintain them. I share those concerns as we are still small team and we need to start having recipes.

Perhaps this is something we can investigate during Oficina time as an improvement to the current strategy of having recipes written by hand.

rafaelramalho19:
We're missing an initial gitignore in the boilerplate 😄

ivolimasilva:
Closing since the open discussions were addressed in #7 .

PR #7, opened 2019-10-08
Feat: folder restructuring with moxy plugins

threequartersjohn:
This PR implements a closer to usable boilerplate suggestion than [its concurrent PR](https://github.com/moxystudio/next-with-moxy/pull/3), with folder restructuring and MOXY plugins.

Further changes to the boilerplate are expected, but this PR hopes to result in something usable asap that can be improved upon as decisions are taken and implemented.


www/shared/data/index.js
```diff
@@ -0,0 +1,2 @@
+// Export Data
```
ivolimasilva: I don't know if this was already discussed but I believe that the _data_ relative to a component should live next to that component. I think there are some use cases where `shared/data` must exist, but generally I believe we could enforce the usage of _data_ files next to its components.

# Proposal

Have files like

```js
.
├─ package.json
├─ package-lock.json
├─ ...
├─ www/
├── pages/
├──── home/
├───── index.js // Export the component here so we can import from 'pages/home'
├───── Home.js // Code of the component
├───── Home.css // Styles of the component
├───── Home.test.js // Test of the component
├───── Home.data.js // Data of the component or it's children if several use it
├─ ...
└─ README.md
```

This means that if some data is used in several places, then it should go up in the tree to the common place (e.g. if a page has two components with the same data, then it's the page that must have that data). If this is not possible then it should live in `shared/data`. With this we should also add rules to the JEST config to ignore _data_ files, so I suggest using those files with `.data.js` extension.

www/app/App.js
```diff
@@ -0,0 +1,66 @@
...
+
+const { publicRuntimeConfig } = getConfig();
+
+const SEO_DATA = {
+    url: 'https://{project-domain}',
+    title: '{project-name}',
+    description: '{project-description}',
+    keywords: ['{project-keyword-1}', '{project-keyword-2}'],
+    image: { src: seoImage, width: 1200, height: 630 },
+};
```
ivolimasilva: Check the comment about _data files_. I suggest placing this properties in a _data file_.

www/pages/home/Home.css
```diff
@@ -0,0 +1,7 @@
+@import "../../shared/styles/variables";
```
ivolimasilva: Check this comment: https://github.com/moxystudio/next-with-moxy/pull/3#discussion_r329499038

www/app/App.js
```diff
@@ -0,0 +1,66 @@
...
+
+class App extends BaseApp {
```
ivolimasilva: I would use `export class App extends NextApp {` for two reasons:

- Calling it `NextApp` instead of `BaseApp` it's easier for someone new to understand we are extending something from Next.js
- Exporting it so we can unit test the component instead of the wrapped component (also we should have tests already)

www/app/App.js
```diff
@@ -0,0 +1,66 @@
...
+    componentDidMount() {
+        keyboardOnlyOutlines();
+
+        // Initialize Google Analytics
+        if (publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID) {
+            ReactGA.initialize(publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID);
+            ReactGA.pageview(window.location.pathname + window.location.search);
+        }
+    }
```
ivolimasilva: I would use the Google Analytics integration like this to ensure it would trigger on every route change:

```js
componentDidMount() {
    keyboardOnly();

    if (NEXT_PUBLIC_GA_TRACKING_ID) {
        ReactGA.initialize(NEXT_PUBLIC_GA_TRACKING_ID);
        ReactGA.pageview(router.asPath);

        this.props.router.events.on('routeChangeComplete', this.handleRouteChange);
    }
}

componentWillUnmount() {
    if (NEXT_PUBLIC_GA_TRACKING_ID) {
        this.props.router.events.off('routeChangeComplete', this.handleRouteChange);
    }
}

// ...

handleRouteChange = (url) => {
    if (NEXT_PUBLIC_GA_TRACKING_ID) {
        ReactGA.pageview(url);
    }
};
```

www/shared/data/index.js
ivolimasilva: Remove `// Export data` comment.

www/shared/styles/base/layout.css
```diff
@@ -0,0 +1,46 @@
...
+   3. Viewport is scalable and occupies at least 320px (iPhone SE)
+*/
+
+html {
+    min-width: 32em;
+    height: 0; /* 3 */
+    min-height: 100%; /* 1 */
+    display: flex; /* 2 */
+    flex-direction: column; /* 2 */
+    -webkit-tap-highlight-color: transparent;
```
ivolimasilva: @satazor @paulobmarcos any idea what this is?

www/pages/home/Home.css
threequartersjohn: I'm not sure if there's been more discussion on this, but i personally find it awkward to import 'imports', which is very nondescript, and would consider a different name. Also, in general, though i agree with not having to hunt for files in imports, i would personally lean towards specificity than generalization. What do you think?

www/app/App.js
```diff
@@ -1,36 +1,41 @@
...
     componentDidMount() {
         keyboardOnlyOutlines();
 
         // Initialize Google Analytics
         if (publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID) {
             ReactGA.initialize(publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID);
-            ReactGA.pageview(window.location.pathname + window.location.search);
+            ReactGA.pageview(this.props.router.asPath);
+
+            this.props.router.events.one('routeChangeComplete', this.handleRouteChange);
```
ivolimasilva: `on`, not `one`.

www/app/App.js
```diff
@@ -1,36 +1,41 @@
...
-import BaseApp from 'next/app';
+import NextApp from 'next/app';
 import Head from 'next/head';
 import keyboardOnlyOutlines from 'keyboard-only-outlines';
 import ReactGA from 'react-ga';
 import getConfig from 'next/config';
-
-import seoImage from '../shared/media/images/seo.png';
 import favicon from '../shared/media/favicons/favicon.ico';
+import SEO_DATA from './App.data';
```
ivolimasilva: I think it's better to have `./App.data.js` since it might seem the file extension is `.data` which is not true. Up for debate, I guess..

www/app/App.js
```diff
@@ -2,40 +2,23 @@ import React from 'react';
...
         keyboardOnlyOutlines();
 
-        // Initialize Google Analytics
-        if (publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID) {
-            ReactGA.initialize(publicRuntimeConfig.NEXT_PUBLIC_GA_TRACKING_ID);
-            ReactGA.pageview(this.props.router.asPath);
-
-            this.props.router.events.on('routeChangeComplete', this.handleRouteChange);
-        }
+        this.unregisterGoogleTracking = registerGoogleTracking(this.props.router);
```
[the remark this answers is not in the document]
satazor: Yep we can always do that later. I will create that task in oficina.

babel.config.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable prefer-import/prefer-import-over-require */
-
 module.exports = (api) => {
     api.cache(false);
```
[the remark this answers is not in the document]
satazor: I actually changed this to `true` as we can cache because we don't have external "dependencies", except from NODE_ENV, but it's stable.

www/app/App.data.js
```diff
@@ -0,0 +1,11 @@
+import seoImage from '../shared/media/images/seo.png';
+
+const SEO_DATA = {
```
ivolimasilva: Use `export default { ...` since each _data_ file should only export one _data_ object.

www/shared/components/contacts/Contacts.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import contacts from './Contacts.data';
```
ivolimasilva: As a convention, I believe we shoud use `.data.js` when importing these types of files so it doesn't seem like `.data` is a valid extension.

postcss.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = require('postcss-preset-moxy')({
+    browsers: require('./package.json').browserslist,
+    url: [
```
satazor: I think we should update this section to match our extensions

PR #12, opened 2019-10-22
feat: implement 3D loader

threequartersjohn:
Implements 3D loader from [this feature update in the plugin](https://github.com/moxystudio/next-common-files/commit/e4a149a9522c4616b82600ed053ff5bd91b4ad48). 

Since there was once a discussion with @satazor about how 3D files would likely always be too large to reasonably send in the bundle, no default `.data-url.` filter was applied.


satazor:
@threequartersjohn 

- Rebased and fixed conflicts
- Fixed package-lock not being updated
- Added `data-url` to all loaders so that they are available for all

PR #13, opened 2019-10-22
fix: fix ignore patterns

threequartersjohn:
Fixes ignore patterns, which, [as per the documentation](https://jestjs.io/docs/en/configuration.html#testpathignorepatterns-array-string), expect regex patterns.

Also ignores `no-useless-escape` rule because it doesn't behave with escaping characters in stringified regex patterns.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -18,8 +18,12 @@ module.exports = {
     // Snapshots
     setupFilesAfterEnv: ['./node_modules/jest-enzyme/lib/index.js'],
     // Coverage
+    collectCoverage: true,
     collectCoverageFrom: [
-        'src/**/*.js',
+        'www/**/*.js',
+    ],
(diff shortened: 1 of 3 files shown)


satazor:
@threequartersjohn I've checked out the branch, tested it and it was still ignoring the data folder. I've made a few changes:

- Changed `collectCoverageFrom` to point to `www` instead of `src`
- Enabled `collectCoverage` by default
- Fixed `coveragePathIgnorePatterns`
- Removed `testPathIgnorePatterns` as it was not necessary

You might want to port these changes to our jest preset. The only thing to be careful is the `www`, which in some projects is `src`. We need to discuss that.

satazor:
Screenshot 2019-10-24 at 10 33 06

The .data file we have is being now correctly ignored.

PR #16, opened 2019-10-25
fix: fix dotenv not being the first require

satazor:


satazor:
Done directly when rebasing here in github

PR #19, opened 2019-11-05
feat: implement moxy jest config

threequartersjohn:
This PR integrates [`jest-config`](https://www.github.com/moxystudio/jest-config) into our boilerplate.


package.json
```diff
@@ -47,6 +47,7 @@
     "react-ga": "^2.6.0"
   },
   "devDependencies": {
```
satazor: I think there are dependencies that we may now remove, such as jest-file.

PR #24, opened 2019-11-19
chore: bump version stylelint-config-moxy

nlfonseca:


www/shared/styles/base/font-faces.css
```diff
@@ -4,9 +4,13 @@
    directory.
    ================================================================================ */
 
+/* stylelint-disable */
```
ivolimasilva: Disable only the rule in question: `/* stylelint-disable scale-unlimited/declaration-strict-value */`

PR #25, opened 2019-11-19
feat: add support for i18n thanks to @moxy/next-intl

satazor:
See https://github.com/moxystudio/next-intl.


satazor:
@ivolimasilva can you take over this PR and add translations to existent static copy, including SEO data?

PR #29, opened 2019-11-26
feat: use runtime configuration when most appropriate

satazor:
Note that this PR disables any static optimization. However, we won't be benefiting from static rendering as we will have intl and stuff that will need `getInitialProps` at the app level.

### DIFF
--- a/www/app/ga-tracking.js
+++ b/www/app/ga-tracking.js
@@ -1,13 +1,16 @@
 import ReactGA from 'react-ga';
+import getConfig from 'next/config';
+
+const { GA_TRACKING_ID } = getConfig().publicRuntimeConfig;
 
 const registerGoogleTracking = (router) => {
-    if (!process.env.GA_TRACKING_ID) {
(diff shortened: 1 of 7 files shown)


satazor:
@threequartersjohn @ivolimasilva  Do we have any section in the docs branch that explains how we should do config? If so, we need to update that as well

satazor:
I believe we must update the docker documentation part if we also have one.

www/app/ga-tracking.js
```diff
@@ -1,13 +1,16 @@
 import ReactGA from 'react-ga';
+import getConfig from 'next/config';
+
+const { GA_TRACKING_ID } = getConfig().publicRuntimeConfig;
```
ivolimasilva: Shouldn't we use `getConfig()` inside the script, making it return the updated value instead of running it when the file is loaded?

satazor:
So we decided to keep using runtime configuration to remain compatible with serverless deployments.

PR #41, opened 2020-01-02
chore: push to current branch after a release

satazor:
The current script posts always to the master branch, which is not always the case. As an example, you might be pushing to a release line branch containing a hotfix.


satazor:
Got the script ready to all this in all our GitHub projects.

Issue #32, opened 2019-12-03
Add tests for the boilerplate components


satazor:
Now that we have a coverage percentage of >= 80%, the CI will fail because of this constraint.

To solve this, we should:

- Implement tests so that coverage is >= 80% (ideally 100%)
- Setup travis & codecov
- Add our typical "build passing" and "coverage"  badges to the README
- Add "Remove `.travis.yml`" to the list of files to delete to our `instructions.md` doc file

ritazoliveira:
Going to close this issue since the changes asked were already made and merged.

PR #44, opened 2020-01-06
docs: disable coverage threshold until we have higher coverage

satazor:
See #32.


satazor:
@ivolimasilva I believe you wanted to copy some of the tests you already have to the boilerplate as part of #32? Meanwhile, we have to disable coverage thresholds so that CI/CD passes.

PR #46, opened 2020-01-07
refactor: import style files

nlfonseca:
now to use variables, custom-medias and mixins we only need to import the "imports":
`@import "../../shared/styles/imports";`

add also the following mixins:
- position
- unstyled-button
- unstyled-list
- layout-main (add a max container width with the responsive adjusts)


www/shared/styles/imports/variables.css
```diff
@@ -0,0 +1,72 @@
...
+:root {
+    --duration-fast: 0.15s;
+    --duration-standard: 0.3s;
+    --duration-long: 0.4s;
+}
+
+/* ==========================================================================
+   Layout vars
+   ========================================================================== */
+:root {
```
fsdiogo: Excuse me for my OCD:

```suggestion

:root {
```

nlfonseca: I love your OCD ❤️

nlfonseca: done!

www/shared/styles/imports/custom-medias.css
```diff
@@ -4,6 +4,9 @@
    Use it like so: @media (--layout-small) { ... }
    ========================================================================== */
 
+/* this need to be according to the sum of --layout-container-width and 2 * --layout-edge-spacing-container-x */
+@custom-media --layout-lte-container-width (width <= 96.25em); /* 1540px */
```
rafaelramalho19: Consider changing this to 
```
@custom-media --layout-lte-container-width (width <= calc(var(--layout-container-width) + calc(2 x  var(--layout-edge-spacing-container-x)) );
```

I haven't tested it tho, just a suggestion 😅 
```

File: www/shared/styles/imports/mixins/layout-main.css [hunk 3]
```diff
@@ -0,0 +1,26 @@
+@import "../variables";
+@import "../custom-medias";
+
+/* Usage: @mixin layout-main .content-selector;;
+============================================= */
```
rafaelramalho19: This divider seems a bit lost here, no? 🤔

www/shared/styles/imports/mixins/unstyled-button.css
```diff
@@ -0,0 +1,13 @@
+@define-mixin unstyled-button {
+    padding: 0;
+    border: none;
+    outline-style: none;
+    background: transparent;
+    border-radius: 0;
+    cursor: pointer;
+    appearance: none;
+
+    &:matches(:active, :focus, :visited, :hover) {
```
rafaelramalho19: This does not work on Edge https://caniuse.com/#feat=css-matches-pseudo

www/shared/styles/imports/variables.css
```diff
@@ -0,0 +1,73 @@
...
+    ============================================= */
+    --layout-edge-spacing-x: 5%;
+
+    /* Header
+    ============================================= */
+    --layout-header-height: 9rem;
+
+    /* Home
+    ============================================= */
+    --layout-home-width: 100vw;
```
rafaelramalho19: I'm not the biggest fan of this solution because you're connecting a specific variable (home) from it's context (`home.css`) into the layout.

Any reason to do this? 🤔

www/shared/styles/imports/custom-medias.css
nlfonseca: `--layout-edge-spacing-container-x` is a percentage. 
And if I'm not mistaken, variables don't work on this file.

rafaelramalho19: @nlfonseca didn't we change `--layout-edge-spacing-container-x` from 6% to 6vh?

rafaelramalho19: That's what I thought, but can you test with the build script and see if the `calc` plugin works? 😄

www/shared/styles/imports/mixins/layout-main.css
fsdiogo: I don't think so, but there is an extra `;` at the end.

www/shared/styles/imports/mixins/unstyled-button.css
nlfonseca: true. this was used when we have cssnext on 
https://cssnext.github.io/features/#matches-pseudo-class
and now we have https://preset-env.cssdb.org/features#matches-pseudo-class
but the moxy/postcss-preset is in stage 3. https://github.com/moxystudio/postcss-preset is 

so this doesn't work 👀

www/shared/styles/imports/variables.css
nlfonseca: no, I'm not a fan either. 
We already have this on the boilerplate. I can remove this! 😃

nlfonseca: done

www/shared/styles/imports/mixins/layout-main.css
nlfonseca: I removed the extra `;`

www/shared/styles/imports/custom-medias.css
nlfonseca: custom-media don't accept variables, so calc doesn't work either.
I will put a warning on these variables to update this custom-media in case of change.

www/shared/styles/imports/custom-medias.css
```diff
@@ -4,6 +4,9 @@
    Use it like so: @media (--layout-small) { ... }
    ========================================================================== */
 
+/* this need to be according to the sum of --layout-container-width and 2 * --layout-edge-spacing-x */
+@custom-media --layout-lte-container-width (width <= 96.25em); /* 1540px */
```
satazor: The `container` seems a bit generic.

We already have `edge-spacing` which means a safe area between the viewport and the actual "app content".  What if we name that container `layout-content`? This way, it would become:

`--layout-content-max-width`
`--layout-content-spacing-x`
`--layout-content-spacing-y`

and finally, this custom media would be `--layout-content-lte-max-width`.

Thoughts?

www/shared/styles/imports/variables.css
```diff
@@ -0,0 +1,68 @@
...
+:root {
+    --color-white: #fff;
+}
+
+/* ==========================================================================
+   Typography
+   ========================================================================== */
+
+:root {
+    --font-montserrat: "Montserrat";
```
satazor: To be honest, I'm not a big fan of this. In my opinion, all typography variants should have names defined in typography and used as mixins. By not having variables here, teams are forced to standardize their typography styles.

See https://github.com/ipfs-shipyard/nomios-web-uikit/blob/master/src/styles/mixins/typography.css for an example of what I mean by standardized typography names.

www/shared/styles/imports/custom-medias.css
nlfonseca: I agree with all less the custom-media name.
I think should be `--layout-lte-content-max-width` to have consistency with the others custom-media.

What you think?

www/shared/styles/imports/variables.css
nlfonseca: The boilerplate already had this variable :P 
But I'm also a fan of having like this. 
Also, I add on `.body` the default font, so I call this variable. 

The same way the teams are forced to change the mixins, here they are forced to change the name of the variable. 
Neither way,  each team needs to standardize their typography styles in each case.

www/shared/styles/imports/custom-medias.css
satazor: Looks good!

nlfonseca: done! 👍

PR #47, opened 2020-01-07
refactor: custom medias to have more breakpoints

nlfonseca:
⚠️ Merge after #46

custom-medias have now more breakpoints.

Besides, this is a new typology, since we are in 2020 and we don't know anymore what is the difference between a tablet and a phone width, I think that we need to be more agnostic.
What do you think about this?

see `www/shared/styles/imports/custom-medias.css`

PR #48, opened 2020-01-07
refactor: typography

nlfonseca:
⚠️ Merge after #47 

Refactor mixins of typography


www/shared/styles/imports/custom-medias.css
```diff
@@ -0,0 +1,44 @@
+/* ==========================================================================
+   Custom media queries
```
rafaelramalho19: Can you add a note:
`> for more informations check https://github.com/postcss/postcss-custom-media` ?

www/shared/styles/imports/mixins/unstyled-button.css
```diff
@@ -0,0 +1,16 @@
...
+    background: transparent;
+    border-radius: 0;
+    cursor: pointer;
+    appearance: none;
+
+    &:active,
+    &:focus,
+    &:visited,
+    &:hover {
+        outline: none;
```
rafaelramalho19: Unpopular opinion...

I would not remove the outline from the focus & active, since it hints keyboards users what is actually selected.

Or at least we should force a "button" component to have the active & focus state declared.

www/shared/styles/imports/variables.css
```diff
@@ -0,0 +1,69 @@
...
+    /* App maximum width - if this is changed you need to update the custom media --layout-lte-container-width
+    ============================================= */
+    --layout-container-width: 140rem;
+
+    /* Spacing around main contents of the website - if this is changed you need to update the custom media --layout-lte-container-width
+    ============================================= */
+    --layout-edge-spacing-x: 5%;
+
+    /* Header
+    ============================================= */
```
rafaelramalho19: I'm not a fan of these dividers. They actually confuse me because it looks like it's separating a previous section from this (and the description is actually above the divider) 😕

www/shared/styles/imports/mixins/unstyled-button.css
fsdiogo: I get your point but as this is an `unstyled-button`, I'd say it's expected for the dev to take care of things like that.

I see this more as of a kind of "button reset", but I can be wrong 🤷‍♂

www/shared/styles/imports/variables.css
fsdiogo: This follows the `idiomatic-css` principles:

https://github.com/necolas/idiomatic-css#3-comments

www/shared/styles/imports/mixins/unstyled-button.css
rafaelramalho19: Removing it from the start makes it easier for the devs to forget totally about it. Just want to raise awareness for accessibility :)

nlfonseca: true, but focus & active need to be treated like a hover. If you are using this mixin, you need to have styles for all states.

www/shared/styles/imports/custom-medias.css
nlfonseca: I will put this on PR !47

www/shared/styles/base/typography.css
```diff
@@ -24,22 +24,33 @@ body {
...
+h4,
+h5,
+h6,
+p {
+    margin: 0;
 }
 
 h1 {
-    @mixin typography-h1 rem;
+    @mixin --type-h1;
```
paulobmarcos: `type` is too generic of a name. `typography` gives better info of what kind of mixin we are using. Also, the file is called `typography` already :D

nlfonseca: agree!

PR #49, opened 2020-01-07
test: increased coverage 

ritazoliveira:
Test coverage was increased to 100%.


intl/messages/test.json
```diff
@@ -0,0 +1,5 @@
+{
```
satazor: It seems strange to have this file here. Can’t we mock it?

package.json
```diff
@@ -37,6 +37,7 @@
     "@moxy/next-compile-node-modules": "^1.3.2",
     "@moxy/next-intl": "^1.1.4",
     "@moxy/next-webpack-oneof": "^1.0.0",
+    "@moxy/next-with-router-ref": "^1.0.0",
```
satazor: It seems all added deps are unnecessary, including this one and the redux ones.

www/app/App.test.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+
+import { App } from './App';
+import { renderWithIntl } from '../shared/test/utils';
+
+beforeAll(() => {
+    document.documentElement.innerHTML = '<div id="__next" style="position: static; top: auto; left: auto; right: auto" />';
```
satazor: Question: why do we need this?

www/app/App.test.js
```diff
@@ -0,0 +1,49 @@
...
+        { ...props } />,
+);
+
+it('should render correctly', () => {
+    const { container } = getTree();
+
+    expect(container).toHaveTextContent('Hello World');
+});
+
+it('should return entries on getInitialProps', async () => {
```
satazor: What does entries mean? Perhaps a better test title would be: `should call page component getInitialProps`? However I don’t see the need to have this test as we don’t have any actual getInitialProps on the app.

www/app/App.test.js
```diff
@@ -0,0 +1,49 @@
...
+                id: 'id',
+                name: 'name',
+            },
+        },
+    });
+
+    expect(result).toMatchObject({ pageProps: 'foo' });
+});
+
+it('should return empty object if not sent component  with getInitialProps on getInitialProps', async () => {
```
satazor: Same as above.

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+const mockedRouter = {
+    asPath: '/foo',
+    events: {
+        on: jest.fn(),
+        off: jest.fn(),
+    },
+};
+
+describe('Google Analytics', () => {
```
satazor: We don’t need a describe here, seems unnecessary.

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+    beforeAll(() => {
+        process.env.GA_TRACKING_ID = '12345';
+    });
+
+    beforeEach(() => {
```
satazor: Maybe call `jest.clearAllMocks()` to simplify?

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+                }),
+            },
+        };
+
+        registerGoogleTracking(router);
+
+        expect(ReactGA.pageview).toHaveBeenCalledTimes(1);
+        expect(ReactGA.pageview).toHaveBeenCalledWith(router.asPath);
+
+        ReactGA.pageview.mockReset();
```
satazor: Change to mockClear.

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+    it('should attach tracking events to router', () => {
+        registerGoogleTracking(mockedRouter);
+
+        expect(mockedRouter.events.on).toHaveBeenCalledTimes(1);
+        expect(mockedRouter.events.on).toHaveBeenCalledWith('routeChangeComplete', expect.any(Function));
+    });
+
+    it('should register a page view if the route changes', () => {
+        let on;
```
satazor: Change to changeHandler?

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+        const router = {
+            ...mockedRouter,
+            events: {
+                ...mockedRouter.events,
+                on: jest.fn((event, handler) => {
```
satazor: We can simplify this. I would leave the mock implementation empty and do a check below, where we both assert that mockedRouter.events.on where called once and with the right args, and then I would grab the handler from .mock.calls[0][1].

www/app/ga-tracking.test.js
```diff
@@ -0,0 +1,75 @@
...
+        on('/bar');
+
+        expect(ReactGA.pageview).toHaveBeenCalledTimes(1);
+        expect(ReactGA.pageview).toHaveBeenCalledWith('/bar');
+    });
+
+    it('should remove tracking events ', () => {
+        registerGoogleTracking(mockedRouter)();
+
+        expect(mockedRouter.events.off).toHaveBeenCalledTimes(1);
```
satazor: Missing assertion to check the event name.

www/shared/test/utils/index.js
```diff
@@ -0,0 +1 @@
+export * from './render-with-intl';
```
satazor: I would rename the folder to test-utils instead of test/utils. Thoughts?

www/shared/test/utils/render-with-intl/render-with-intl.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
satazor: We don’t need to test the test utils.. instead I would ignore the test utils folder altogether in our jest config. There’s an option for this, see https://jestjs.io/docs/en/configuration#coveragepathignorepatterns-arraystring.

Moreover, we can then simplify the file structure and remove the folder as it will contain just the actual file.

www/shared/test/utils/render-with-intl/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as renderWithIntl } from './render-with-intl';
```
satazor: This export is unnecessary. However, we will remove this file altogether, please continue reading.

satazor:
@ritazoliveira As part of this PR, can you manually revert the changes made with this commit? https://github.com/moxystudio/next-with-moxy/commit/11376e270cc5593d14d913d70a98e8e05910f0da

www/shared/test/utils/render-with-intl/render-with-intl.test.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    it('should correctly rerender', () => {
+        const render = renderWithIntl(
+            <FormattedMessage id={ 'foo' } />,
+        );
+
+        render.rerenderWithIntl('bar');
+
+        expect(render.container.innerHTML).toMatchSnapshot();
```
threequartersjohn: Could you please use `getByText` here as well? I personally think we should avoid snapshots if the content is small, which it is in this case. 

Also, i would suggest destructuring `container`, `renderWithIntl` et al, like the test above. I think it looks prettier and easier to read!

www/shared/test/utils/render-with-intl/render-with-intl.test.js
threequartersjohn: Could you please expand on why testing this is unnecessary? I'd agree if this were an external library, or otherwise code that is likely tested elsewhere, but this tool is of our own creation, and i think this should be tested to establish trust in this util.

intl/messages/test.json
ritazoliveira: Yes, makes more sense, done!

www/app/App.test.js
ritazoliveira: I agree, this test and the next one may be deleted. Thanks!

www/shared/test/utils/render-with-intl/render-with-intl.test.js
ritazoliveira: Will do! Thanks!

ritazoliveira: Will do! Thanks.

package.json
ritazoliveira: Yes, some were added while migrating the tests but in the end they were not needed.

www/shared/test/utils/render-with-intl/render-with-intl.test.js
threequartersjohn: After conversation with @satazor, he agrees with keeping these tests.

ritazoliveira:
Requested changes were applied. @PedroMiguelSS maybe you can review it now.

intl/messages/test.json
satazor: @ritazoliveira I think you forgot to remove this file.

package.json
satazor: I think you forgot to handle this as well.

package.json
```diff
@@ -49,7 +50,8 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-ga": "^2.6.0",
-    "react-intl": "^3.6.0"
+    "react-intl": "^3.6.0",
+    "redux": "^4.0.5"
```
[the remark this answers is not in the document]
satazor: They were deleted meanwhile.

www/shared/tests/utils/render-with-intl/render-with-intl.test.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import { renderWithIntl } from './render-with-intl';
+
+it('should correctly render formatted message', () => {
+    const { container, getByText } = renderWithIntl(
+        <FormattedMessage id={ 'foo' } />,
```
satazor: Can we change to `id="foo"`?

www/shared/tests/utils/render-with-intl/render-with-intl.test.js
```diff
@@ -0,0 +1,23 @@
...
+    );
+
+    expect(container.innerHTML).toMatchSnapshot();
+    expect(getByText('foo')).toBeInTheDocument();
+});
+
+it('should correctly rerender with passed children', () => {
+    const { rerenderWithIntl, getByText } = renderWithIntl(
+        <FormattedMessage id={ 'foo' } />,
```
satazor: Can we change to `id="foo"`?

www/document/Document.test.js
```diff
@@ -0,0 +1,25 @@
...
+    Html: ({ children }) => children,
+    Head: ({ children }) => <head>{ children }</head>, // eslint-disable-line react/prop-types
+    Main: () => <div />,
+    NextScript: () => <div />,
+    default: require('react').Component,
+}));
+
+import { Document } from './Document';
+
+beforeEach(() => {
```
PedroMiguelSS: We can replace `beforeEach` by `beforeAll` because we don't need to inject `__NEXT_INTL_POLYFILL_URL__` variable before each test. In this test suite, this is not that important as we only have one test and `beforeEach` will run only once. But we should replace it so we can avoid this problem in the future.

www/shared/tests/utils/render-with-intl/render-with-intl.test.js
```diff
@@ -0,0 +1,23 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import { renderWithIntl } from './render-with-intl';
+
+it('should correctly render formatted message', () => {
+    const { container, getByText } = renderWithIntl(
+        <FormattedMessage id="foo" />,
+    );
+
+    expect(container.innerHTML).toMatchSnapshot();
```
PedroMiguelSS: IMO this assertion is unnecessary. 

`expect(getByText('foo')).toBeInTheDocument();` is enough to get 100% coverage.

PR #54, opened 2020-01-10
fix: rerender with intl test

threequartersjohn:
Fixes `rerenderWithIntl` test. Missed this during code review, so here's the PR with the fix.

As it was, this test was not testing whether `rerenderWithIntl` was wrapping the `intl` provider, because no child would require it.

With a `FormattedMessage` child, this test now actually tests if `rerenderWithIntl` wraps its children with the necessary provider and does not throw any errors.


satazor:
Looks good.

PR #57, opened 2020-01-17
feat: nextjs v9.2.0 and add CSS support with css modules and postcss

ritazoliveira:
Updated Next.js version to [9.2.0](https://nextjs.org/blog/next-9-2) which introduces **built-In CSS Support** for both global stylesheets and component-level styles. 
The component-level stylesheets now have a `module.css `extension so that CSS Modules are used and global stylesheets were moved to a` styles/global `folder. 
Some packages needed to be updated to accommodate the changes.


satazor:
I've tested this in own of our recent projects and it's working great. Here's the before and after network requests:

Screenshot 2020-01-17 at 12 11 20
Screenshot 2020-01-17 at 12 17 24

satazor:
@ritazoliveira the PR looks good but we must update the docs.

- Change the CSS modules page in "whats-included"
- Change file structure to reflect `.module.css`

www/shared/styles/mixins/typography.css
```diff
@@ -3,7 +3,6 @@
 /* ==========================================================================
   Fonts
   ========================================================================== */
-
 @define-mixin font-montserrat $fontWeight: 400 {
```
satazor: Can you restore this line?

PR #58, opened 2020-01-23
docs: add splash screen recipe

satazor:
Also:

- Update deps
- Standardized recipe ids

PR #59, opened 2020-01-23
test: add AppTree component to be used in tests

satazor:
The `renderWithIntl` test util has some design flaws:

- It doesn't scale when adding other providers
- It doesn't allow using renderToString (useful to test SSR) as it uses RTL render method internally

This change adds a `AppTree` test component instead that setups all the app providers.
It contains just the `IntlProvider` but developers may add more as their app grows.

Moreover, the `IntlProvider` messages now use an identity object.


intl/messages/en-US.json
```diff
@@ -1,5 +1,8 @@
 {
     "home.title": "Homepage",
+
     "contacts.name": "Name: { name }",
-    "contacts.email": "Email: { email }"
+    "contacts.email": "Email: { email }",
+
+    "splash.error": "Oops, something went bad. Please refresh to try again."
```
ivolimasilva: I believe this message is related to https://github.com/moxystudio/next-with-moxy/pull/58 and not to this PR.

```suggestion
```

satazor:
Thanks for the review @ivolimasilva!

PR #61, opened 2020-01-30
chore: changed google analytics library and updated respective tests

ritazoliveira:
Changed the library used to integrate the app with google analytics and updated the tests from `ReactGA` to [`ganalytics`](https://github.com/lukeed/ganalytics).
Also updated the `App.js `component to be a stateless component.


www/app/App.js
```diff
@@ -1,54 +1,56 @@
...
-                </Head>
-                <KeyboardOnlyOutlines>
-                    <Component { ...pageProps } />
-                </KeyboardOnlyOutlines>
-            </>
-        );
-    }
-}
+App.propTypes = {
+    Component: PropTypes.elementType,
```
satazor: Shouldn't all these props be required?

www/shared/utils/google-analytics.js
```diff
@@ -0,0 +1,25 @@
+import ganalytics from 'ganalytics';
+
+const ga = typeof window !== 'undefined' && process.env.GA_TRACKING_ID ? ganalytics(process.env.GA_TRACKING_ID, {}, true) : () => {};
```
satazor: Shouldn't the mock `ga` be an object with `send: () => {}`?

www/shared/utils/google-analytics.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const trackOnRouteChanged = (router) => {
+    const handleRouteChange = () => ga.send('pageview');
+
+    if (!ga.send) {
```
satazor: See my comment above.

www/app/App.js
```diff
@@ -1,54 +1,56 @@
...
-import registerGoogleTracking from './ga-tracking';
+import { trackOnRouteChanged } from '../shared/utils';
 import favicon from '../shared/media/favicons/favicon.ico';
 import SEO_DATA from './App.data.js';
 
-export class App extends NextApp {
-    componentDidMount() {
-        this.unregisterGoogleTracking = registerGoogleTracking(this.props.router);
-    }
+const App = ({ Component, pageProps, router }) => {
```
satazor:
```suggestion
export const App = ({ Component, pageProps, router }) => {
```

Tests are failing because we are not exporting the app here.

www/app/App.js
ritazoliveira: Yeah, they should! Fixed it!

www/shared/utils/google-analytics.js
```diff
@@ -0,0 +1,22 @@
+import ganalytics from 'ganalytics';
+const ga = typeof window !== 'undefined' && process.env.GA_TRACKING_ID ?
```
satazor: Add a new line

www/shared/utils/google-analytics.test.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+beforeEach(() => {
+    jest.resetModules();
+    jest.clearAllMocks();
+
+    process.env.GA_TRACKING_ID = '12345';
+});
+
+it('should do nothing if env var is not defined', () => {
```
satazor: I think all this tests are associado with `trackRouteChanged`, therefore we should use a `describe`. Furthermore, we are missing tests for the exported `ga`: test that ga works with or without the env car

www/app/App.js
```diff
@@ -1,54 +1,56 @@
...
+import { trackOnRouteChanged } from '../shared/utils';
 import favicon from '../shared/media/favicons/favicon.ico';
 import SEO_DATA from './App.data.js';
 
-export class App extends NextApp {
-    componentDidMount() {
-        this.unregisterGoogleTracking = registerGoogleTracking(this.props.router);
-    }
+export const App = ({ Component, pageProps, router }) => {
+    useEffect(() => {
```
satazor: This can be simplified to just `useEffect(() => trackRouteChanged(router))`

PR #64, opened 2020-02-12
feat(ci): Fully qualified URL for SEO image

jgradim:
Add SITE_URL environment variable to replace `{project-domain}` placeholder

Example application: https://frontend-mr1-4izymcemfa-ew.a.run.app/
Using the above URL in twitter and facebook card validators should show the default placeholder image and values

Twitter
---

Screen Shot 2020-02-12 at 14 29 23

Facebook
---

Screen Shot 2020-02-12 at 14 29 55

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,3 +5,6 @@
 
 # Google Analytics Tracking Code
 #GA_TRACKING_ID=
+
+# Fully qualified site URL, without trailing slash
+#SITE_URL=http://localhost:3000
--- a/.env.sample
+++ b/.env.sample
@@ -7,4 +7,5 @@
 #GA_TRACKING_ID=
 
 # Fully qualified site URL, without trailing slash
-#SITE_URL=http://localhost:3000
+# This variable is used in www/app/App.data.js to have absolute paths in SEO images
+SITE_URL=http://localhost:3000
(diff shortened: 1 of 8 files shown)


.env.sample
```diff
@@ -5,3 +5,6 @@
 
 # Google Analytics Tracking Code
 #GA_TRACKING_ID=
+
+# Fully qualified site URL, without trailing slash
```
satazor: It would be useful to complement with an example where this is used

.env.sample
```diff
@@ -5,3 +5,6 @@
...
+#SITE_URL=http://localhost:3000
```
satazor: Maybe we should actually consider this as the default value, so I would say to uncomment it.

docusaurus/docs/welcome/instructions.md
```diff
@@ -21,10 +21,7 @@ Keep in mind, part of adapting this boilerplate into a deliverable project is al
     - `Contacts.data.js`: In the `name` field. This file is a sample file and you might want to delete it altogether.
     - In this document, in case you want to use it.
 
-2. Change all instances of `{project-domain}` into the domain that will be used in production:
-
-    - `package.json`: In the `description` field.
-    - `App.data.js`: In the `url` field.
+2. For local development, create a `.env` file and set a value for `SITE_URL` environment variable. You will likely want to use the provided example, `http://localhost:3000`
```
satazor: If you apply my suggestion above, we actually don't need this step, right?

jgradim: We still need to copy `.env.sample` to `.env` regardless of wether the value is preset or not. Adjusting instructions.

PR #70, opened 2020-03-13
chore: update eslint config

dominguesgm:
Given the recent refactoring of [MOXY's eslint-config](https://github.com/moxystudio/eslint-config), this PR switches the `eslint` configuration used from the previous deprecated one to the current version.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,12 +1,8 @@
 {
     "root": true,
     "extends": [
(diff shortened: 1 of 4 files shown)


satazor:
@dominguesgm I think we might need to update the eslint and stylelint page in the documentation (docusaurus). I believe it references the old package by its name.

PR #75, opened 2020-03-19
chore: update to new eslint-config

satazor:
This PR updates to the new eslint-config with a11y enabled as well. There's no warnings besides these ones:

Screenshot 2020-03-19 at 20 26 12

These are related a11y + Next.js Link API. Associated threads:

- https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/402
- https://github.com/zeit/next.js/issues/5533

What should we do? Ignore warnings? Or implement a custom Link to deal with this issue?


satazor:
@acostalima we might want to do the same for rnwm.

threequartersjohn:
I think we might explicitly unset the rule in our `eslintrc.json`. I don't think this is a problem with the plugin, but **Next.js**'s implementation. This also makes it easy to rollback if the `Link` component is changed to have a different API.

Even custom `Link` components will likely set `href` conditionally, which would also fire this warning. Also, in **Next.js**'s `Link` component the `href` is a required prop, which already safeguards against the presence of an `href`attribute.

satazor:
@threequartersjohn let me elaborate on the custom link component: we could do a custom Link component in src/shared/components that would use a anchor by default and would also support a custom render prop in cases we want to render anything else. The warning would still be there, but we would ignore (with eslint-ignore) just in that file.

I also like disabling the rule. Let’s ask for feedback from @moxystudio/developers.

satazor:
Disabled the rule for now.

PR #76, opened 2020-03-22
fix: validate SITE_URL env var as a valid URL

threequartersjohn:
Implements validation to guarantee that `SITE_URL` is a valid URL. Currently, it only allows `http` for `localhost`, which might be too strict, so opinions welcome.

This implementation makes it so failure to validate will stop the app building process. 

This implementation also parallels the [`next-validation`](https://github.com/moxystudio/next-validation) package, which is made obsolete for this specific problem with this solution.

### DIFF
--- a/docusaurus/docs/what-is-included/environment-variables.md
+++ b/docusaurus/docs/what-is-included/environment-variables.md
@@ -51,6 +51,19 @@ if (process.env.FEATURE_A) {
     }
     ```
 
+    If you want to validate environment variables, we suggest you do it in this step. We suggest using [`@hapi/joi`](https://github.com/hapijs/joi) with its [`attempt`](https://github.com/hapijs/joi/blob/master/API.md#attemptvalue-schema-message-options) function, like the following example:
+    ```js
+    {
+        env: {
+            // ...
+            FEATURE_A: Joi.attempt(process.env.FEATURE_A, Joi.boolean()
+                .truthy('1')
+                .falsy('0')
+                .message('process.env.FEATURE_A must be one of "true", "1", "false", or "0".')),
+        },
+    }
+    ```
+
 3. Add it to the Dockerfile
 
     ```dockerfile
--- a/docusaurus/docs/what-is-included/environment-variables.md
+++ b/docusaurus/docs/what-is-included/environment-variables.md
@@ -51,7 +51,8 @@ if (process.env.FEATURE_A) {
     }
     ```
 
-    If you want to validate environment variables, we suggest you do it in this step. We suggest using [`@hapi/joi`](https://github.com/hapijs/joi) with its [`attempt`](https://github.com/hapijs/joi/blob/master/API.md#attemptvalue-schema-message-options) function, like the following example:
+    If you want to validate environment variables, we suggest you to use [`@hapi/joi`](https://github.com/hapijs/joi) with its [`attempt`](https://github.com/hapijs/joi/blob/master/API.md#attemptvalue-schema-message-options) function, like the following example:
+
     ```js
     {
         env: {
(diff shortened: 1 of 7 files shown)


next.config.js
```diff
@@ -59,6 +60,10 @@ module.exports = (phase, nextConfig) =>
         compress: process.env.COMPRESSION !== '0',
         env: {
             GA_TRACKING_ID: process.env.GA_TRACKING_ID,
-            SITE_URL: process.env.SITE_URL,
+            SITE_URL: Joi.attempt(process.env.SITE_URL,
+                Joi.string()
+                    .uri({ scheme: ['https', /http.*localhost/] })
```
satazor: Why `/http.*localhost/`? Wasn't `http` sufficient?

satazor:
Looks good!

next.config.js
threequartersjohn: It was, but this is to ensure  HTTP is only allowed for `localhost`.

satazor: I see! Well it's surprising that it works as `scheme` refers to the part before the `:`.

satazor: Maybe we can use `try`?

```
Joi.alternatives().try(
    Joi.string().uri({ scheme: ['https'] }).pattern(/\/$/, { invert: true })
    Joi.pattern(/http:\/\/localhost:\d+/),
).message('process.env.SITE_URL must be a valid URL')),
```

Haven't tested.

threequartersjohn: > I see! Well it's surprising that it works as `scheme` refers to the part before the `:`.

Oh, that's not in the documentation 🤔 hmm yes, it's working now because it's actually looking for the last instante of `:`, and because i was testing with `localhost:3000` it was working...

threequartersjohn: We could change it, but i would also like some opinions about restricting it to `https`. I'm not sure if this is wise, i think it's ok to expect of development team to be able to use `https` but i'm not sure it's ok to expect other stakeholders to do so.

satazor: I'm fine blocking it to `https` and only `http` for development. I'm just unsure if this is the right place to validate or do these kind of enforcement.

I would KISS and not enforce it as this level.

docusaurus/docs/what-is-included/environment-variables.md
```diff
@@ -51,6 +51,16 @@ if (process.env.FEATURE_A) {
     }
     ```
 
+    If you want to validate environment variables, we suggest you do it in this step. We suggest using [`@hapi/joi`](https://github.com/hapijs/joi) with its [`attempt`](https://github.com/hapijs/joi/blob/master/API.md#attemptvalue-schema-message-options) function, like the following example:
+    ```js
+    {
+        env: {
+            // ...
+            FEATURE_A: Joi.attempt(process.env.FEATURE_A, Joi.boolean().message('process.env.FEATURE_A must have a boolean value.')),
```
satazor:
```suggestion
            FEATURE_A: Joi.attempt(process.env.FEATURE_A, Joi.boolean()
                   .truthy('1'').falsy('0')
                   .message('process.env.FEATURE_A must be one of "true", "1", "false" or "0".')
             ),
```

Just wanted to complement with a bit more realistic example, as `1` and `0` are often used as booleans in env variables. Feel free to adjust indentation.

next.config.js
threequartersjohn: Understood, will remove it for now.

threequartersjohn:
Many thanks to @satazor for handling deploys on `now` 👏

satazor:
@threequartersjohn thanks, I'm still tinkering with it to comment on PRs and to deploy correctly on master

satazor:
Merging! Let's see if it works on master.

PR #86, opened 2020-04-03
fix: typography target media

AfonsoVReis:
Right now `typography.css` isn't using custom medias accordingly to what we set in `custom-medias.css`.

This PR aims to fix this.


satazor:
Thanks @AfonsoVReis, nice catch!

PR #87, opened 2020-04-13
docs: contentful recipe

dominguesgm:
Include recipe documentation for using the Contentful CMS.

Contains instructions on how to configure Contentful's content models, integration on the application side, and more advanced details like usage with Redux, preview banner rendering, translations, custom per-page SEO and custom caching.


docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,411 @@
...
+
+A CMS, or Content Management System, is a platform that helps in the creation and management of content to be consumed in a web application. In short, these services provide a database and a database management system with a layer of usability on top that makes them inclusive to users without technical expertise.
+
+### Contentful
+
+[Contenful](https://www.contentful.com/) is one such CMS, and the object of this recipe.
+
+For more information about **Contentful**'s API please [refer to their documentation](https://www.contentful.com/developers/docs/).
+
+## Modeling your database
```
acostalima, line 19: I believe the correct terminology is schema and not database.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,411 @@
...
+
+    // Create default client
+    let client = createClient({
+        space: process.env.CONTENTFUL_SPACE_ID,
+        accessToken: process.env.CONTENTFUL_TOKEN, // Delivery API Token
+        host: 'cdn.contentful.com', // Delivery API host
+    });
+
+    // Check whether route has `cms-preview` query parameter
+    const isPreviewingContentful = Object.hasOwnProperty.call(router.query, 'cms-preview');
```
acostalima, line 121: Where does this `cms-preview` property come from exactly? By query parameter do you mean an actual URL query parameter like `?cms-preview=true`?

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,411 @@
...
+This function also has the advantage letting you return props to your App function, making it easier to signal the App itself if you're in preview mode.
+
+```js
+// App.js
+import { createClient } from 'contentful';
+
+App.getInitialProps = async ({Component, context, router}) => {
+
+    // Create default client
+    let client = createClient({
```
acostalima, line 114: Defer the creation of the client to when we know whether we should create a client pointing to a live or preview.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,411 @@
...
+        id: 'en-US',
+        name: 'English',
+        loadMessages: async () => {
+            const content = await client.getEntries({
+                content_type: 'genericTranslations',
+                locale: 'en-US',
+                // The parameter bellow is needed to avoid the problem described at the top of the file.
+            });
+
+            return content.items[0].fields; melhorar isto
```
acostalima, line 336: Do you still need to work on this?

docusaurus/docs/recipes/contentful.md
acostalima: Nevermind, I completely missed this in the documentation. 😄

dominguesgm:
Some issues came up in regards to the Custom Cache Layer which I would like to discuss.

In short, we need to know the `SITE_URL` in order to setup the Contentful client to direct its requests there so our endpoint can serve as a proxy to the Contentful server (this only applies in production environments).

However, when we have PR/MR preview, the `SITE_URL` is not correctly configured to that preview's URL and the application is still launched with `NODE_ENV` set to `production`. This will cause the Contentful client to point to a non existing or non accessible endpoint.

Should we perhaps use a different `NODE_ENV` value in staging servers so as to avoid proxying the Contentful requests in those scenarios? I don't know if there is a way to obtain the preview url and set an environment variable with it before launching the preview.

Any thoughts on this @satazor @acostalima @AfonsoVReis @paulobmarcos?

satazor:
@dominguesgm the SITE_URL is being correctly configured for staging and for preview environments. @jgradim has updated the gitlab yml of the CI/CD pipeline to have that.

Are you referring to GitLab or GitHub?

docusaurus/docs/recipes/contentful.md
AfonsoVReis: actually these are my leftovers :D @dominguesgm i'm sorry i didn't remove this from here on the handoff, please remove now 🙏

satazor:
@dominguesgm I think we should edit the pipeline recipe to use `moxy.in` in SITE_URL for staging (releases + latest master),

dominguesgm:
I can create an issue so we can tackle it after, sounds good?

satazor:
@dominguesgm sounds good to me.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+
+### Contentful
+
+[Contenful](https://www.contentful.com/) is one such CMS, and the object of this recipe.
+
+For more information about **Contentful**'s API please [refer to their documentation](https://www.contentful.com/developers/docs/).
+
+## Modeling your schema
+
+As a developer working on the App that will consume the content stored in the CMS, you're also responsible for creating and modeling the database itself. This will mean creating new content models that must be usable by the client team which will eventually handle **Contentful** after the hand off of the project. One **major** consideration to have is the usability and readability of the content models you're creating. Remember that they are meant to be usable without any previous technical knowledge of this context.
```
jdiogopeixoto, line 21: > modeling the database

Change to "modeling the schema"?

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+
+As a developer working on the App that will consume the content stored in the CMS, you're also responsible for creating and modeling the database itself. This will mean creating new content models that must be usable by the client team which will eventually handle **Contentful** after the hand off of the project. One **major** consideration to have is the usability and readability of the content models you're creating. Remember that they are meant to be usable without any previous technical knowledge of this context. 
+
+**Contentful** offers many options for validations and input appearance and descriptions to help the users understand what kind of input is expected, and you should use all of them to minimize the chance of users not understanding what's expected of them, in what ways they are limited, reducing the chances of unexpected content in your application.
+
+Keep in mind:
+
+→ Always use appropriate types when possible,
+
+→ Always use validation where appropriate,
```
jdiogopeixoto, line 29:
```suggestion
→ Always use validations where appropriate,
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+
+API Rate limits specify the number of requests a client can make to Contentful APIs in a specific time frame. Every request counts against a per second rate limit.
+
+Currently Conentful doesn't enforce any limits on requests that hit their CDN cache. For requests that do hit the Content Delivery API enforces rate limits of 78 requests per second.
+
+When a client gets rate limited, the API responds with the `429 Too Many Requests` status code and sets the `X-Contentful-RateLimit-Reset` header that tells the client when it can make its next request. 
+
+### 2. Custom Caching Layer
+
+One preventive measure for avoiding hitting the rate limit for Contentful is to implement your own custom caching layer.
```
jdiogopeixoto, line 348: > One preventive measure for avoiding hitting the rate

This sentence doesn't sound good. I would change to "One preventive measure to avoid hitting the rate".

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+const seoData = {
+    title: contentfulSEO.title || defaultSEO.title,
+    meta: [...defaultSEO.meta, ...contentfulSEO.meta],
+    link: [...defaultSEO.link, ...contentfulSEO.link],
+}
+```
+
+- Render the SEO data (title and tags) inside the `Head` in the `App.js` file.
+
+For the rendering step, we recommend using [@moxy/next-seo](https://www.npmjs.com/package/@moxy/next-seo), as this package takes care of two things:
```
jdiogopeixoto, line 509: I don't think this should be only a recommendation as we already use the package in the boilerplate, I would enforce the use of `next-seo`. Also the point of creating the parser is exactly to transform the information to be used in the `Seo` component. Thoughts?

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+        title: seoContent.title,
+        meta: seoContent.meta.map((entry) => mapValues(entry, (value) => getContent(value, contentfulData.includes.Asset)))
+        link: seoContent.link.map((entry) => mapValues(entry, (value) => getContent(value, contentfulData.includes.Asset)))
+    }
+
+    return content;
+}
+```
+
+- Merge the _default_ SEO data with the SEO data fetched from the Contentful.
```
jdiogopeixoto, line 497: Is this necessary? The `Seo` component already does this.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,514 @@
...
+        return asset.fields.file.url.slice(2);
+    }
+
+    return content;
+};
+
+// Provided with the result from the Contentful API call and the entryID for the current page content,
+// This function will extract the SEO data, and complete it with the URL of any needed asset.
+
+const contentfulSEOParser = (contentfulData, entryID) => {
```
jdiogopeixoto, line 472: Can we name it `parseContentfulSEO` or something similar?

docusaurus/docs/recipes/contentful.md
dominguesgm: I don't have a personal preference, so sure, I'll change it

docusaurus/docs/recipes/contentful.md
dominguesgm: Makes sense, we can change the wording from "recommendation" to "use `next-seo`".

docusaurus/docs/recipes/contentful.md
dominguesgm: The `SEO` component only accepts one object with SEO data. If you have local SEO data and SEO data from Contentful, you need to combine it in the object you will provide to the `SEO` component.

One could argue the defaults could be written directly to `Head` in the `App` component, right before the `SEO` component is rendered, but this way you wouldn't be taking advantage of the tag de-duplication behavior that the `SEO` component provides.

docusaurus/docs/recipes/contentful.md
dominguesgm: Yeah, I think it's a good idea to have it similar in both recipes 👍

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
+---
+id: contentful
+title: Implementing Contentful
+sidebar_label: Contentful
+---
+
+This recipe aims to guide you through the implementation of Contetful in the project and also provides custom solutions for different use-cases.
```
fsdiogo, line 7:
```suggestion
This recipe aims to guide you through the implementation of Contentful in the project and also provide custom solutions for different use-cases.  
```

docusaurus/docs/recipes/contentful.md
acostalima: Generally, considering this is a recipe for Contentful, I'd say we should adhere to Contentful's where relevant. I've suggested "schema" instead of "database", but is "schema" the term adopted by Contentful?

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+[Contenful](https://www.contentful.com/) is one such CMS, and the object of this recipe.
+
+For more information about **Contentful**'s API please [refer to their documentation](https://www.contentful.com/developers/docs/).
+
+## Modeling the schema
+
+As a developer working on the App that will consume the content stored in the CMS, you're also responsible for creating and modeling the schema itself. This will mean creating new content models that must be usable by the client team which will eventually handle **Contentful** after the hand off of the project. One **major** consideration to have is the usability and readability of the content models you're creating. Remember that they are meant to be usable without any previous technical knowledge of this context. 
+
+**Contentful** offers many options for validations and input appearance and descriptions to help the users understand what kind of input is expected, and you should use all of them to minimize the chance of users not understanding what's expected of them, in what ways they are limited, reducing the chances of unexpected content in your application.
```
paulobmarcos, line 23:
```suggestion
**Contentful** offers many options for type validations, input appearance and field descriptions to help the users understand what kind of input is expected. You should use all of them to minimize the chance of users not understanding what's expected of them, in what ways they are limited, reducing the chances of unexpected content in your application.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+Keep in mind:
+
+→ Always use appropriate types when possible,
+
+→ Always use validations where appropriate,
+
+→ Write titles and descriptions to ensure the user knows what's expected of that field.
```
paulobmarcos, line 31:
```suggestion
→ Use meaningful  titles and descriptions to ensure the user knows what's expected of that field.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+As a developer working on the App that will consume the content stored in the CMS, you're also responsible for creating and modeling the schema itself. This will mean creating new content models that must be usable by the client team which will eventually handle **Contentful** after the hand off of the project. One **major** consideration to have is the usability and readability of the content models you're creating. Remember that they are meant to be usable without any previous technical knowledge of this context. 
+
+**Contentful** offers many options for validations and input appearance and descriptions to help the users understand what kind of input is expected, and you should use all of them to minimize the chance of users not understanding what's expected of them, in what ways they are limited, reducing the chances of unexpected content in your application.
+
+Keep in mind:
+
+→ Always use appropriate types when possible,
+
+→ Always use validations where appropriate,
```
paulobmarcos, line 29:
```suggestion
→ Use validations where appropriate to make sure the input values are what the developer expects,
```

docusaurus/docs/recipes/contentful.md
jdiogopeixoto: Looks like Contentful calls it [data model](https://www.contentful.com/developers/docs/concepts/data-model/). But I think schema is understandable from a developer's point of view.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+## Localization
+
+**Contentful** makes it easy to add localized content to your application. Once you define a locale, which you can do in the Locales option in the Settings menu, you can specify in each field you create whether they can be localized, in what languages, etc., giving you fine control over what needs to be localized and what doesn't. 
+
+Once having these locales, you can then use the **Contentful** API signaling your desired locale through the `locale` field. You'll find an example of this in the next section.
+
+## Preview
+
+Contentful allows you to fetch content that is not in a `published` state, also including content that is in a `changed` or `draft` state.  This can help users see how content would look on their website before publishing it and making it available on for every visitor.
```
paulobmarcos, line 41:
```suggestion
Contentful splits content changes in 3 states:
- `Published` : when a user saves some content.
- `Changed`: when the user changes previously saved content but still hasn't saved the new one.
- `Draft`: when the user creates new content but doesn't save.

Usually, only `Published` content is available, unless a specific request is done. This request is called `Preview` and may help users see beforehand how content would look like on their website before publishing it and making it available on for every visitor.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+### 1. Installing Contentful SDK
+
+Using Contenful in your application can be done using [the **Contentful** SDK available as an npm package](https://www.npmjs.com/package/contentful), which you can install using the following command:
+
+```sh
+npm install --save contenful
+```
+
+This SDK provides access to **Contentful**'s [Delivery AP](https://www.contentful.com/developers/docs/references/content-delivery-api/)I and [Preview API](https://www.contentful.com/developers/docs/references/content-preview-api/), both of which will be used in the integration of **Contentful** in your app.
```
paulobmarcos, line 55:
```suggestion
This SDK provides access to **Contentful**'s [Delivery API](https://www.contentful.com/developers/docs/references/content-delivery-api/) and [Preview API](https://www.contentful.com/developers/docs/references/content-preview-api/), both of which will be used in the integration of **Contentful** in your app. 
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+Using Contenful in your application can be done using [the **Contentful** SDK available as an npm package](https://www.npmjs.com/package/contentful), which you can install using the following command:
+
+```sh
+npm install --save contenful
+```
+
+This SDK provides access to **Contentful**'s [Delivery AP](https://www.contentful.com/developers/docs/references/content-delivery-api/)I and [Preview API](https://www.contentful.com/developers/docs/references/content-preview-api/), both of which will be used in the integration of **Contentful** in your app. 
+
+First, you'll need **Contentful**'s access tokens, of which you'll need three. These can all be found in the Settings menu, under the `API keys` option:
```
paulobmarcos, line 57:
```suggestion
First, you'll need two **Contentful**'s access tokens and a space identifier. All of the following can be found in the Settings menu, under the `API keys` option:
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+        },
+    },
+],
+```
+
+## Cache
+
+### 1. Contentful API rate limits
+
+API Rate limits specify the number of requests a client can make to Contentful APIs in a specific time frame. Every request counts against a per second rate limit.
```
paulobmarcos, line 340:
```suggestion
API Rate limits specify the maximum number of requests a client can make to Contentful APIs in a specific time frame. Every request counts against a per second rate limit.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,513 @@
...
+
+> ❗️ The `process.env.SITE_URL` variable has to be correctly configured and accessible from both server and client side, otherwise the request to the proxy endpoint will not happen correctly. This may mean that preview url's in merge request will not function correctly, which will happen if the application is started with `process.env.NODE_ENV` set to `production`.
+
+## Custom SEO
+
+There may be cases where you will want to configure custom SEO per page. Unfortunately, Contentful does not provide out-of-the-box SEO support, so you will need to implement your own strategy. The approach we suggest is the following:
+
+- Create a content model for SEO.
+- Add a field `Title` that should be only used to identify the model, e.g. "Homepage SEO".
+- Add a SEO field (not-rich long text) to the model. Here, the SEO related tags (title, meta, etc.) will be defined in a _json_ format like so:
```
paulobmarcos, line 413:
```suggestion
- Add a SEO field (json type) to the model. Here, the SEO related tags (title, meta, etc.) will be defined in a _json_ format like so:
```

docusaurus/docs/recipes/contentful.md
paulobmarcos: The SEO component uses the `Head` component internally. If we use it with the defaults in the `App` component, only the new ones set by the page will be overridden. The previous default ones that are not overridden will still be available. So there is no need to do this I think 🤔

docusaurus/docs/recipes/contentful.md
dominguesgm: Misspell fixed, but I think `provides` is more correct

docusaurus/docs/recipes/contentful.md
dominguesgm: I understand the motivation behind the suggesting, and "not-rich long text" should be changed. But the reasoning behind this was to indicate the type of input (Contentful-wise) you have to create.

Perhaps just leaving as "long text"?

dominguesgm:
I wanted to write this as a response to the last comment in your review, @paulobmarcos, but I don't seem to be able to.

What if you have a default tag for `og:image` and then write one through the `SEO` component? You wouldn't be able to use keys to avoid duplication as the default `og:image` tag would be set before the `SEO` component is rendered.

This is why the approach I suggested was combining the default and CMS tags in an object that will be fed to the `SEO` component.

docusaurus/docs/recipes/contentful.md
paulobmarcos: imagem

Contentful-wise is `JSON object` :+1:

dominguesgm: 🤦 didn't see it

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,520 @@
...
+
+## Preview
+
+Contentful splits content changes in 3 states:
+
+- `Published` : when a user saves some content.
+- `Changed`: when the user changes previously saved content but still hasn't saved the new one.
+- `Draft`: when the user creates new content but doesn't save.
+  
+Usually, only `Published` content is available, unless a specific request is done. This request is called `Preview` and may help users see beforehand how content would look like on their website before publishing it and making it available on for every visitor.
```
acostalima, line 47:
```suggestion
Usually, only `Published` content is available, unless a specific request is done. This request is called `Preview` and may help users see beforehand how content would look like on their website before publishing it and making it available for every visitor.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,520 @@
...
+```js
+const result = await client.getEntries({
+    content_type: 'clients',
+    locale: 'en-US'
+});
+```
+
+### 2. Populating the app with fetched content
+
+To populate the app with the fetched content, we will explore two possible implementations: one for bigger applications that justify the inclusion of Redux store, and those which don't.
```
acostalima, line 105: Maybe I'm missing something, but the English sounded a bit strange to me. 🤔 

```suggestion
To populate the app with the fetched content, we will explore two possible implementations: one for bigger applications that justify the inclusion of a Redux store and other which doesn't.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,520 @@
...
+
+Using hooks: 
+
+```js
+// App.js
+const App = ({ Component, pageProps, rootSelector, router }) => {
+    const [isPreviewingContentful, setIsPreviewingContentful] = useState(false);
+    
+    // Using a useEffect hook with no dependencies guaranties that it fires only once per instance of App
+    useEffect(() => setIsPreviewingContentful(Object.hasOwnProperty.call(router.query, 'cms-preview')), [])
```
acostalima, line 299:
```suggestion
    useEffect(() => setIsPreviewingContentful(Object.hasOwnProperty.call(router.query, 'cms-preview')), [router, setIsPreviewingContentful])
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,520 @@
...
+],
+```
+
+## Cache
+
+### 1. Contentful API rate limits
+
+API Rate limits specify the maximum number of requests a client can make to Contentful APIs in a specific time frame. Every request counts against a per second rate limit.
+
+Currently Conentful doesn't enforce any limits on requests that hit their CDN cache. For requests that do hit the Content Delivery API enforces rate limits of 78 requests per second.
```
acostalima, line 349:
```suggestion
Currently, Contentful doesn't enforce any limits on requests that hit their CDN cache. For requests that do hit, the Content Delivery API enforces rate limits of 78 requests per second.
```

docusaurus/docs/recipes/contentful.md
threequartersjohn: @acostalima, as mentioned in the comment, this `useEffect` must have no dependencies so that it doesn't memoize any object that can change, this in effect guarantees that it only fires once.

This does break one of our current eslint rules though.

Also, huge thanks @dominguesgm for working on my mistakes 🙏

acostalima: Which comment? Sorry, I'm kind of lost. 🙏

acostalima: We can ignore `setIsPreviewingContentful` since its reference never changes anyway.

acostalima: Can `router` change after App is mounted?

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+
+    const { contentfulClient } = context;
+
+    // Fetch data from Contentful
+    const entries = await contentfulClient.getEntries({
+        content_type: 'useCase',
+    });
+
+    // Return result as props
+    return { entries }
```
satazor, line 162: Missing ;

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+To populate the app with the fetched content, we will explore two possible implementations: one for bigger applications that justify the inclusion of a Redux store, and another which doesn't.
+
+#### 2.1. Accessing the client directly though `getInitialProps`
+
+In Next.js apps, the `getInitialProps` function in general components typically receives a context object, which exposes many aspects of the context in which the page is rendered. Because we are building a custom app, and doing the work of running these `getInitialProps` functions ourselves, we can control what data they receive. In practice, we will be populating the context object with the information we want to propagate to our components. So, in your `App.js` file, you can check whether the page was loaded with a `cms-preview` query parameter, build your client, and share it across the app.
+
+This function also has the advantage letting you return props to your App function, making it easier to signal the App itself if you're in preview mode.
+
+```js
+// App.js
```
satazor, line 114: This path should be the full like src/app/App.js. All other code blocks suffer from the same problem.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+In the following example we will be checking the presence of a `cms-preview` query parameter to decide whether to show preview content or the standard, published content, i.e. something like `www.example.com/?cms-preview`. 
+
+```js
+// buildStore.js
+import { createClient } form 'contentful';
+import { applyMiddleware, createStore } from 'redux';
+import thunkMiddleware from 'redux-thunk';
+
+export const buildStore = (initialState, { query }) => {
+
```
satazor, line 183: Remove this empty line.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+    const reducer = {
+        // Example reducer
+    };
+    
+    // Add client as a thunk middleware
+    // This will give us access to the `client` object in our action functions as an argument
+    const middlewares = applyMiddleware(thunkMiddleware.withExtraArgument({ client }));
+
+    return createStore(reducer, initialState, middlewares);
+    }
```
satazor, line 214: Missing ;

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+export default buildStore;
+```
+
+Then, in your actions scripts, you can use the provided client to fetch the content that you need, like so:
+
+```js
+// usecases/actions.js
+// Create a fetch action that fetches data from Contentful
+const fetch = () => async (dispatch, getState, { client }) => {
+    dispatch({ type: 'Fetching'})
```
satazor, line 226: There’s a lot of missing ; in this file. Please check all other files as well.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+        });
+    }
+};
+```
+
+### 3. Conditionally rendering DOM elements in case of preview
+
+Since we're on the topic, you must use a similar strategy in your `App.js` file to conditionally render a DOM element to let the user know they are in a preview state. This must happen once per instance of the App, since we will not perpetuate the `cms-preview` query parameter on route changes.
+
+Class component:
```
satazor, line 263: I would put the hook version first as our boilerplate is no longer using the class version.

satazor: Moreover, next.js even removed the custom App using class components, so I don’t know if extending is the correct way here.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+// App.js
+const App = ({ Component, pageProps, rootSelector, router }) => {
+    const [isPreviewingContentful, setIsPreviewingContentful] = useState(false);
+    
+    // Using a useEffect hook with no dependencies guaranties that it fires only once per instance of App
+    useEffect(() => setIsPreviewingContentful(Object.hasOwnProperty.call(router.query, 'cms-preview')), []);
+
+    return (
+        (...)
+            { this.isPreviewingContentful && <ContentfulPreview> }
```
satazor, line 303: `this` Is not valid here.

Missing / in the JSX tag of ContentfulPreview. Maybe we should add a note stating that it’s a component that would render a visible warning on the site saying we are on preview mode, like a ribbon. This is valid for the dato CMS recipe as well.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+if (process.env.NODE_ENV === 'production') {
+    clientOptions.host = process.env.SITE_URL;
+    clientOptions.basePath = '/api/contentful';
+}
+
+const client = createClient(clientOptions);
+```
+
+**NOTES:**
+> ❗️ You have to be able to access to `process.env.NODE_ENV` from the client side as well as the server side, to proxy client side requests to Contentful in production environments.
```
satazor, line 410: Nifty pick: always add new line after `:`, check other places too.

Moreover I would use the warning emoji if applicable to keep consistency with other recipes.

docusaurus/docs/recipes/contentful.md
satazor: Also there’s a missing space.

docusaurus/docs/recipes/contentful.md
```diff
@@ -0,0 +1,511 @@
...
+    } catch (parsingError) {
+        console.err('Unvalid JSON in SEO content');
+        console.err(parsingError);
+        return;
+    }
+
+    // Convert every content: { id: ''} into a content: url, keep unchanged otherwise
+    const content = {
+        title: seoContent.title,
+        meta: seoContent.meta.map((entry) => mapValues(entry, (value) => getContent(value, contentfulData.includes.Asset)))
```
satazor, line 497: Missing coma at the end of the line.

I would pass all files through the longer, there are a lot of linting and JS errors in the code blocks.

docusaurus/docs/recipes/contentful.md
dominguesgm: Should we perhaps remove the Class component approach then?

satazor: Yep I think it’s better.

docusaurus/docs/recipes/contentful.md
threequartersjohn: @acostalima, you are correct, we can set `router` and `setIsrPreviewingContentful`  as dependencies. My original intent was for developers not to memoize `router.query` specifically, which auto-fix in `eslint` does automatically (i.e., if you leave the dependency array empty, and run auto-fix, `eslint` will set `router.query` as a dependency).

I think the documentation should reflect this, but i'm willing to do a PR for this after this first merge if more discussion is warranted.

acostalima: I've never came across a situation where I was using an `useEffect` referencing props or state without specifying its deps. In this case, as a general rule, we should always specify the deps as best practices suggest.

PR #89, opened 2020-04-15
feat: add next-seo

jdiogopeixoto:
This PR adds [next-seo](https://github.com/moxystudio/next-seo) to the boilerplate.

### DIFF
--- a/www/app/App.data.js
+++ b/www/app/App.data.js
@@ -1,9 +1,37 @@
 import seoImage from '../shared/media/images/seo.png';
 
-export default {
+export const data = {
     url: process.env.SITE_URL,
     title: '{project-name}',
     description: '{project-description}',
-    keywords: ['{project-keyword-1}', '{project-keyword-2}'],
+    keywords: [
+        '{project-keyword-1}',
+        '{project-keyword-2}',
+    ],
     image: { src: `${process.env.SITE_URL}${seoImage}`, width: 1200, height: 630 },
 };
+
+export const seoData = {
+    title: data.title,
+    meta: [
+        { name: 'description', content: data.description },
+        { name: 'keywords', content: data.keywords.join(' ') },
+        /* Facebook & search engines */
+        { property: 'og:url', content: data.url },
+        { property: 'og:type', content: 'website' },
+        { property: 'og:title', content: data.title },
+        { property: 'og:description', content: data.description },
+        { property: 'og:site_name', content: data.title },
+        { property: 'og:image', content: data.image.src },
+        { property: 'og:image:width', content: data.image.width },
+        { property: 'og:image:height', content: data.image.height },
+        /* Twitter */
+        { name: 'twitter:card', content: 'summary_large_image' },
+        { name: 'twitter:url', content: data.url },
+        { name: 'twitter:title', content: data.title },
+        { name: 'twitter:description', content: data.description },
+        { name: 'twitter:image', content: data.image.src },
+    ],
+};
+
+export default data;
--- a/www/app/App.js
+++ b/www/app/App.js
@@ -4,11 +4,12 @@ import Head from 'next/head';
 import KeyboardOnlyOutlines from '@moxy/react-keyboard-only-outlines';
 import { withNextIntlSetup } from '@moxy/next-intl';
 import { LayoutTree } from '@moxy/next-layout';
+import Seo from '@moxy/next-seo';
 import { CookiesProvider } from 'react-cookie';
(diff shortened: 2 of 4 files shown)


www/app/App.js
```diff
@@ -36,22 +34,8 @@ export const App = ({ Component, pageProps }) => {
...
-                <meta property="og:image:width" content={ SEO_DATA.image.width } />
-                <meta property="og:image:height" content={ SEO_DATA.image.height } />
-                {/* Twitter */}
-                <meta property="twitter:card" content="summary_large_image" />
-                <meta property="twitter:url" content={ SEO_DATA.url } />
-                <meta property="twitter:title" content={ SEO_DATA.title } />
-                <meta property="twitter:description" content={ SEO_DATA.description } />
-                <meta property="twitter:image" content={ SEO_DATA.image.src } />
             </Head>
+            <Seo data={ seoData } />
```
dominguesgm, line 38: If I'm not mistaken the `Seo` component should still be placed inside `Head`. Personally, I'd prefer to have it right after the opening `Head` tag

jdiogopeixoto: It renders the `Head` tag so it shouldn't be inside it: https://github.com/moxystudio/next-seo/blob/master/src/Seo.js#L35

www/app/App.js
```diff
@@ -36,22 +34,8 @@ export const App = ({ Component, pageProps }) => {
                 <meta name="msapplication-TileColor" content="#ffffff" />
                 <meta name="msapplication-config" content="/favicons/browserconfig.xml?v=M4KN2GElyG" />
                 <meta name="theme-color" content="#ffffff" />
```
paulobmarcos, line 36: I know the above is specific to favicons and it might not directly be "seo", but maybe we could still use the Seo component to deal with this? 🤷

jdiogopeixoto: That is a good point but I think we should keep it there and let me explain why.

We have instructions on [how to setup the favicon](https://next-with.moxy.tech/docs/welcome/instructions/#3-setup-favicon--manifest) and we recommend the use of [RealFaviconGenerator](https://realfavicongenerator.net/). This generator returns a similar HTML to what we have right now, so people just have to replace it. By moving that out it would be necessary to change it in two files, which is not hard but it's an extra step. Thoughts?

paulobmarcos: Yes, the change would require that. Is not too much of an hassle since the usage of the Seo component is pretty straight forward and we would have a single place to manage these. Would like some other opinions on this.

nlfonseca: I agree with @paulobmarcos

www/app/App.js
nlfonseca, line 38: if we want to change this on a specific page, we only need to send a new object for data to `Seo` component?
the information will be merged with the default?

www/app/App.data.js
```diff
@@ -1,9 +1,37 @@
...
+    meta: [
+        { name: 'description', content: data.description },
+        { name: 'keywords', content: data.keywords.join(' ') },
+        /* Facebook & search engines */
+        { property: 'og:url', content: data.url },
+        { property: 'og:type', content: 'website' },
+        { property: 'og:title', content: data.title },
+        { property: 'og:description', content: data.description },
+        { property: 'og:site_name', content: data.title },
+        { property: 'og:image', content: data.image },
```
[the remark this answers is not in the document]
jdiogopeixoto: Nice catch, I fixed it. Thanks 👍

www/app/App.js
jdiogopeixoto: You would have to do the same, import the `Seo` and use it. 
All the meta tags that already exist will be replaced by the ones you pass.

nlfonseca: Nice!

www/app/App.js
paulobmarcos: I think we can do this change then :+1:

jdiogopeixoto: I changed it and updated the instructions. The meta tags related to the favicon are now being handled by the `Seo` component.
Let me know your thoughts.

paulobmarcos: After speaking with Andre, we decided to let the meta and links tags related to the icons as it was.

Sorry for the extra work :+1: Can we roll back?

nlfonseca: so, what do you think to do a component like `<Seo>` for favicons?

```
<Head>
    <Favicons data={ faviconsData } />
    <Seo data={ seoData } />
</Head>
```

PR #90, opened 2020-04-21
fix: contentful recipe improvements

dominguesgm:
Implemented fixes mentioned in #87 after the PR was merged, along a few other fixes.

Introduced details on what environment variables to setup.


jdiogopeixoto:
I noticed that you are using `src/...` in some code blocks. Shouldn't it be `www/...`?

dominguesgm:
Nice catch @jdiogopeixoto, you're right, changing it

docusaurus/docs/recipes/contentful.md
```diff
@@ -111,7 +119,7 @@ In Next.js apps, the `getInitialProps` function in general components typically
 This function also has the advantage letting you return props to your App function, making it easier to signal the App itself if you're in preview mode.
```
paulobmarcos, line 119:
```suggestion
This function also has the advantage of letting you return props to your App function, making it easier to signal the App itself if you're in preview mode.
```

docusaurus/docs/recipes/contentful.md
```diff
@@ -174,13 +182,12 @@ Given the nature of Next.js and the implementation of [next-redux-wrapper](https
 In the following example we will be checking the presence of a `cms-preview` query parameter to decide whether to show preview content or the standard, published content, i.e. something like `www.example.com/?cms-preview`. 
 
 ```js
-// buildStore.js
-import { createClient } form 'contentful';
+// www/shared/utils/build-store/buildStore.js
```
paulobmarcos, line 185: I'm not sure if this is still the right path based on our redux recipe: https://next-with-moxy-recipes.moxy.tech/docs/recipes/setting-up-redux

docusaurus/docs/recipes/contentful.md
```diff
@@ -207,30 +214,29 @@ export const buildStore = (initialState, { query }) => {
...
 export default buildStore;
 ```
 
-Then, in your actions scripts, you can use the provided client to fetch the content that you need, like so:
+Then, in your fetcher scripts, you can use the provided client to fetch the content that you need, like so:
 
 ```js
-// usecases/actions.js
-// Create a fetch action that fetches data from Contentful
+// www/shared/utils/fetchers/useCases.js
```
paulobmarcos, line 229: Same as above, please confirm the path is in conformity with our reduxe recipe.

dominguesgm: I see the problem here, I replaced this while unaware of the redux context, I'm reverting to what it was before

docusaurus/docs/recipes/contentful.md
```diff
@@ -260,13 +266,33 @@ const fetch = () => async (dispatch, getState, { client }) => {
...
+}
+```
+
 Class component:
 
 ```js
-// App.js
-class App extends Nextapp {
+// www/app/App.js
+class App extends NextApp {
```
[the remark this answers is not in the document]
dominguesgm: Did Next drop the support for App class component entirely or just started recommending a  function  App component? I could not find info on this on the documentation.

If it was just a change in recommended approach, I think we should keep it. If support is being dropped in future versions, I agree we should remove it.

PR #91, opened 2020-04-21
docs: improve datocms recipe

jdiogopeixoto:
This PR aims to improve the DatoCMS recipe with two particularly relevant changes:
- Adds the [`next-with-apollo`](https://www.npmjs.com/package/next-with-apollo) package;
- Changes Apollo client cache to rely only on the network.


docusaurus/static/recipes-assets/dato-cms/next-with-apollo/next-with-apollo.test.js
```diff
@@ -0,0 +1,41 @@
...
+beforeEach(() => {
+    process.env.DATOCMS_TOKEN = 'foo';
+});
+
+afterEach(() => {
+    jest.resetModules();
+});
+
+it('should render correctly', () => {
+    const withApollo = require('./next-with-apollo').default;
```
paulobmarcos, line 14: Can withApollo be imported at the top?

paulobmarcos: Seems like it runs only once if we import it at the top, right @jdiogopeixoto?  Resolving this.

PR #94, opened 2020-05-07
feat: add react-player recipe

jdiogopeixoto:
This PR adds a recipe on how to use `react-player` with some common use cases.

### DIFF
--- a/docusaurus/docs/recipes/video-player.md
+++ b/docusaurus/docs/recipes/video-player.md
@@ -1,0 +1,103 @@
+---
+id: video-player
+title: Adding a video player
+sidebar_label: Video player
+---
+
+Having videos in your website is a common need. In this recipe we will show you how to use [`react-player`](https://github.com/CookPete/react-player), a package that helps adding video content to your website, along with some common use cases.
+
+## Walk-through
+
+### 1. Install `react-player`
+
+```sh
+npm i react-player
+```
+
+### 2. Use `ReactPlayer` in your project
+
+You just need to import `ReactPlayer` and start using it.
+
+```js
+// ...
+import ReactPlayer from 'react-player';
+
+const Home = () => (
+    <div className={ styles.home }>
+        <ReactPlayer url="https://www.video-url.com" />
+    </div>
+);
+```
+
+### 3. Common use cases
+
+Now let's go through some common use cases that might be helpful.
+
+> ⚠️ By default [`react-player`](https://github.com/CookPete/react-player) sets a width of 640px and a height of 360px to the video player. You should change it using [`width` and `height` props](https://github.com/CookPete/react-player#props) in order to make it [responsive](https://github.com/CookPete/react-player#responsive-player) according to your needs.
+
+#### 1. Autoplay
+
+To enable autoplay you need to use the [`playing` prop](https://github.com/CookPete/react-player#props), that lets you control the playing of the video. It can be particularly useful when you have a loading state and you only want the video to play after the loading is over.
+
+```js
+const [isPlaying, setIsPlaying] = useState(true);
+
+// ...
+
+<ReactPlayer url="https://www.video-url.com" playing={ isPlaying } />
+```
+
+ℹ️ Be sure to check the different browser policies for [autoplaying videos](https://github.com/CookPete/react-player#autoplay) as they are different and you might need additional configuration. For example, on Chrome you might need to mute the video in order to autoplay it, or in Safari you might need additional work to make it autoplay on low power mode devices which is a common problem.
+
+#### 2. Enable/disable default controls
+
+To enable or disable the video controls you need to use the [`controls` prop](https://github.com/CookPete/react-player#props).
+
+```js
+<ReactPlayer url="https://www.video-url.com" controls />
+```
+
+#### 3. Disable download
+
+The download of videos is enabled by default. To disable it you need to change the [`controlsList` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList) using the [`config` prop](https://github.com/CookPete/react-player#props).
+
+```js
+<ReactPlayer
+    url="https://www.video-url.com"
+    config={ {
+        file: {
+            attributes: {
+                controlsList: "nodownload"
+            }
+        }
+    } } />
+```
+
+#### 4. Disable context menu
+
+The context menu is enabled by default. To disable it you need to change the [`onContextMenu` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) and prevent the default behavior.
+
+```js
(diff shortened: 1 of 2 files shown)


docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+
+To enable or disable the video controls you can use the `controls` prop.
+
+```js
+<ReactPlayer url="https://www.video-url.com" controls />
+```
+
+#### 3. Disable download
+
+The download of videos is enable by default. To disable it you need to change the [`controlsList`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList) using the `config` prop.
```
threequartersjohn, line 58: Typo, should be `enabled`.

Also, i would write it `need do change the controlList property, using the...`, i.e. adding property after mentioning the property. This might seem like nickpicking but it incentivizes maintaining terminology across documentations, which keeps things clearer and traceable for the reader.

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+import ReactPlayer from 'react-player';
+
+const Home = () => (
+    <div className={ styles.home }>
+        <ReactPlayer url="https://www.video-url.com" />
+    </div>
+);
+```
+
+### 3. Usage
```
threequartersjohn, line 32: Not sure this is a good title for use cases, `usage` is a broad term that does not specify any particular user need. For example, consider how `Usage` is interchangeable with the `Walk-through` title in this document. I would maybe name this `Common use cases` to be as specific to this section as possible.

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+
+Now let's go through some common use cases that might be helpful.
+
+> ⚠️ By default `react-player` sets a width of 640px and a height of 360px to the player. You should change it in order to fit your needs.
```
threequartersjohn, line 36: If this warning is necessary, it should also be noted that height and width are changed through props (`height` and `width`).

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### 1. Autoplay
+
+To enable autoplay you need to use `playing` prop.
+
+```js
+<ReactPlayer url="https://www.video-url.com" playing />
+```
+
+Be sure to check the different browser policies for autoplaying videos as they are different and you might need additional configuration. For example on Chrome you might need to mute the video in order to autoplay it.
```
threequartersjohn, line 46: Mentioning Chrome is ok, but we should definitely mention Safari as well, which is a fair bit more inconsistent and cumbersome with video, especially in low power mode mobile devices.

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+sidebar_label: Video player
+---
+
+It's often normal to have the necessity of having videos in your website. In this recipe we will show you how to use [`react-player`](https://github.com/CookPete/react-player) and some common use cases.
+
+## Walk-through
+
+### 1. Install `react-player`
+
+```sh
```
threequartersjohn, line 13: Not entirely sure this is necessary, but i would add a `--save` arg here.

jdiogopeixoto: I did check that and it's not necessary. It adds `react-player` to the dependencies with or without the `--save`.

docusaurus/docs/recipes/video-player.md
jdiogopeixoto: Sure. Can you give more info on that? I am not familiar with those problems.

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+        }
+    } } />
+```
+
+#### 4. Disable context menu
+
+The context menu is enabled by default. To disable it you need to change the [`onContextMenu` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) and prevent the default behavior.
+
+```js
+const handleOnContextMenu = useCallback((event) => {
```
PedroMiguelSS, line 77:
```suggestion
const handleContextMenu = useCallback((event) => {
```

docusaurus/docs/recipes/video-player.md
```diff
@@ -0,0 +1,99 @@
...
+    event.preventDefault();
+}, []);
+
+// ...
+
+<ReactPlayer url="https://www.video-url.com" onContextMenu={ handleOnContextMenu } />
```
PedroMiguelSS, line 83:
```suggestion
<ReactPlayer url="https://www.video-url.com" onContextMenu={ handleContextMenu } />
```

docusaurus/docs/recipes/video-player.md
threequartersjohn: Sure, in Webkit when a device is in low power mode autoplay seems to be inconsistently disabled. You can find [examples here](https://stackoverflow.com/questions/50400902/detect-if-ios11-device-is-in-low-power-mode-to-prevent-bad-ux-on-normally-correc) of people struggling with autoplay in low power mode. This is compounded by there being no good way to verify if a device is in low power mode or if the video stopped playing. In an earlier project i hacked up a solution that pools the video's ready state until enough data to play continuously, or give up after an interval, with a condition like so:

```js
videoRef.current.getInternalPlayer().readyState >= HTMLMediaElement.HAVE_CURRENT_DATA
```

I don't think these are worth mentioning since they're fairly obscure and hacky, but they're just an example of struggling with Webkit because of obscure or undocumented behaviors.

docusaurus/docs/recipes/video-player.md
```diff
@@ -29,33 +29,37 @@ const Home = () => (
...
-<ReactPlayer url="https://www.video-url.com" playing />
+const [isPlaying, setIsPlaying] = useState(true);
+
+// ...
+
+<ReactPlayer url="https://www.video-url.com" playing={ isPlaying } />
 ```
 
-Be sure to check the different browser policies for autoplaying videos as they are different and you might need additional configuration. For example on Chrome you might need to mute the video in order to autoplay it.
+ℹ️ Be sure to check the different browser policies for [autoplaying videos](https://github.com/CookPete/react-player#autoplay) as they are different and you might need additional configuration. For example, on Chrome you might need to mute the video in order to autoplay it, or in Safari you might need additional work to make it autoplay on low power devices which is a common problem.
```
threequartersjohn, line 50: Just a small nitpick, it's not exactly low power devices, but `low power mode`, which is a battery saving mode in iPhones. 

Otherwise, lgtm 👍

PR #99, opened 2020-06-02
fix: fix vercel deployment

satazor:


acostalima:
@satazor, can we close this?

satazor:
I still need this to fix CD on Zeit, something is causing it to fail.

PR #101, opened 2020-06-18
chore: add custom render

guilhermedeandrade:
This PR will add a `customRender` function which is able to wrap a desired component into all the necessary providers (and anything else it needs). Note that this approach is recommended by the RTL's [documentation](https://testing-library.com/docs/react-testing-library/setup#custom-render).

I also took this opportunity to refactor some tests which were using the `container` property from a `render` instead of `screen` ([as suggested here](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen))

Motivation: https://github.com/moxystudio/react-native-with-moxy/pull/42#pullrequestreview-432694206

### DIFF
--- a/www/shared/test-utils/modules/react-app-tree/AppTree.test.js
+++ b/www/shared/test-utils/modules/react-app-tree/AppTree.test.js
@@ -1,26 +1,26 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import AppTree from './AppTree';
 
 it('should render children correctly', () => {
-    const { container } = render(
+    render(
         <AppTree>
             <div>foo</div>
         </AppTree>,
     );
 
-    expect(container).toHaveTextContent('foo');
+    expect(screen.getByText('foo')).toBeInTheDocument();
 });
 
 it('should correctly setup IntlProvider', () => {
-    const { container } = render(
+    render(
         <AppTree>
             <FormattedMessage id="foo" />
         </AppTree>,
     );
 
-    expect(container).toHaveTextContent('foo');
+    expect(screen.getByText('foo')).toBeInTheDocument();
 });
 
 it('should correctly setup IntlProvider with overrides', () => {
@@ -28,11 +28,11 @@ it('should correctly setup IntlProvider with overrides', () => {
         messages: { foo: 'bar' },
     };
 
-    const { container } = render(
+    render(
         <AppTree intlProvider={ intlProvider }>
             <FormattedMessage id="foo" />
         </AppTree>,
     );
 
-    expect(container).toHaveTextContent('bar');
+    expect(screen.getByText('bar')).toBeInTheDocument();
 });
--- a/docusaurus/static/recipes-assets/react-splash-screen/SplashScreen.test.js
+++ b/docusaurus/static/recipes-assets/react-splash-screen/SplashScreen.test.js
@@ -1,50 +1,49 @@
 import React from 'react';
-import { render } from '@testing-library/react';
 import WaitForReact from '@moxy/react-wait-for-react';
 import SplashScreen from './SplashScreen';
-import AppTree from '../../test-utils/modules/react-app-tree';
+import { render, screen } from 'test-utils';
 
 jest.mock('@moxy/react-wait-for-react', () => jest.fn(() => null));
 
 it('should render the content and progress bar correctly when progress is 0', () => {
     WaitForReact.mockImplementationOnce(({ children }) => children({ progress: 0, error: undefined }));
 
-    const { container } = render(<AppTree><SplashScreen /></AppTree>);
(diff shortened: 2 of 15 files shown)


www/shared/test-utils/modules/react-app-tree/AppTree.test.js
```diff
@@ -1,38 +1,38 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import AppTree from './AppTree';
 
 it('should render children correctly', () => {
-    const { container } = render(
+    render(
```
satazor, line 7: Instead of using screen, shouldn’t we use the container returned by render?

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
```diff
@@ -1,38 +1,38 @@
...
         </AppTree>,
     );
 
-    expect(container).toHaveTextContent('foo');
+    expect(screen.getByText('foo')).toBeInTheDocument();
 });
 
 it('should correctly setup IntlProvider', () => {
-    const { container } = render(
+    render(
```
satazor, line 17: I think we are rendering AppTree inside another, because the custom render uses AppTree as well. Maybe, for all tests in this file, we should pass the options.wrapper equal to a simple div?

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
satazor: I saw your comment now. Maybe we should then be using screen in all other tests or you were having issues with it?

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
guilhermedeandrade: Oh yes, we are! I wil do this adjustment. Thanks for pointing that out 🙌

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
guilhermedeandrade: I think I'm already using `screen` whatever it's possible to do it, but I'll double check the test files anyway.

guilhermedeandrade: Maybe you find strange that `container` is still being used quite often. However, I can't replace it with `screen` if the assertion needs to use `querySelector` for instance or anything else from the DOM API.

guilhermedeandrade: Just double checked and there are still cases where `container` isn't needed at all. Example: https://github.com/moxystudio/next-with-moxy/pull/101/files#diff-46c595aa0102a72440ca2a3bf2d1bd5a

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
guilhermedeandrade: Actually, this is using the render from the library itself, not the custom one. Should we keep it as it is or pass the wrapper option as you suggested?

satazor: It’s good as is!

www/shared/test-utils/modules/react-app-tree/AppTree.test.js
satazor: All good then?

guilhermedeandrade: All good 🙌

acostalima: I've opened an [issue](https://github.com/moxystudio/eslint-config/issues/101) recently in our ESLint config to consider adding two new plugins for RTL. One of them has a [rule](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-screen-queries.md) precisely enforce the usage of screen.

acostalima: > Maybe you find strange that `container` is still being used quite often. However, I can't replace it with `screen` if the assertion needs to use `querySelector` for instance or anything else from the DOM API.

True. However, perhaps we can create a few [custom queries](https://testing-library.com/docs/react-testing-library/setup#add-custom-queries) to find and select elements with custom attributes:

- `[data-wait-for-react-element="progressBar"]`
- `[data-wait-for-react-element="splashScreen"]`

guilhermedeandrade: Good idea @acostalima! I will add a custom query for this case

guilhermedeandrade: Done @acostalima

acostalima: Good to go! 🚀

PR #102, opened 2020-06-19
fix: fix bold font weights in firefox drawing bolder than specified

threequartersjohn:
Firefox sometimes draws fonts bolder than specified in styles rules. This is an old issue but still seems to be happening, for example:

- without the fix:
image example of no fix

- whereas it should be:
Screenshot 2020-06-19 at 12 52 48

Further information can be found in [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=857142) discussing the implementation of this CSS rule.

### DIFF
--- a/www/shared/styles/global/typography.css
+++ b/www/shared/styles/global/typography.css
@@ -13,6 +13,7 @@ body {
     @mixin text-smooth;
 
     color: var(--color-black);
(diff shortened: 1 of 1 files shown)


satazor:
The `text-smooth` mixin already has that style. Maybe you were not using it in your project?

PR #113, opened 2020-08-26
feat: upgrade to use @moxy/next-router-scroll

### LINKED ISSUE #112: Update recipe assets tests
As they are using the old app tree utils and so on.

satazor:
Also:

- Added a reusable PageSwapper component to avoid duplication.
- Use RTL best practices, including the new user-event package.
- Update recipes to use new test utils.

Closes #112


AfonsoVReis (approved):
👍

PR #115, opened 2020-10-08
chore: add user-event

guilhermedeandrade:
This PR will add [user-event](https://testing-library.com/docs/ecosystem-user-event) which is a companion library for Testing Library that provides more advanced simulation of browser interactions than the built-in fireEvent method.

Example:

```js
import { screen } from '@testing-library/dom'
import userEvent from '@testing-library/user-event'

test('types inside textarea', async () => {
  document.body.innerHTML = `<textarea />`

  await userEvent.type(screen.getByRole('textbox'), 'Hello, World!')
  expect(screen.getByRole('textbox')).toHaveValue('Hello, World!')
})
```

Note that using `fireEvent` the syntax for changing an input value isn't as readable:
`fireEvent.change(input, { target: { value: 'foo' } })`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4306,9 +4306,9 @@
       }
     },
     "@testing-library/user-event": {
-      "version": "12.1.3",
-      "resolved": "https://registry.npmjs.org/@testing-library/user-event/-/user-event-12.1.3.tgz",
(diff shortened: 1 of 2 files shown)


acostalima (approved):
This PR is not adding `user-event` but just updating the library's version. Anyhow, LGTM, I guess. 😉

guilhermedeandrade:
> This PR is not adding `user-event` but just updating the library's version. Anyhow, LGTM, I guess. 😉

The project which I'm working on doesn't have it, so I thought the template didn't 😂 

Should have checked, sorry.

PR #116, opened 2020-11-20
chore(cookies): fix plural wording

threequartersjohn:
Changes all instances of `cookieConsents` and `CookieConsents` to `cookiesConsent` and `CookiesConsent`, respectively, where the change in wording is meant to express "The user provides **consent** for **cookies**".

This is because consent is an uncountable noun in this context. `Consents` is a valid word, but only when the consent in question relates to e.g. documents or otherwise numerable items that individually express some consent, which is not the case here.

### DIFF
--- a/www/app/App.js
+++ b/www/app/App.js
@@ -18,8 +18,8 @@ import { seoData } from './App.data.js';
 export const AppInner = ({ Component, pageProps }) => {
     const pageKey = usePageKey();
 
-    const handleCookieConsents = useCallback((cookieConsents) => {
-        if (cookieConsents.includes('analytics')) {
+    const handleCookiesConsent = useCallback((cookiesConsent) => {
+        if (cookiesConsent.includes('analytics')) {
             initGTM();
         } else {
             destroyGTM();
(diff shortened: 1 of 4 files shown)


satazor:
The module currently only has one consent, which is analytics, but is prepared to allow for other use cases as well. Therefore there are multiple consents.

Having that said, shouldn’t we be using plural as it’s a collection of consent?

satazor:
Note that cookies is just the way the consents are stored, we could be using local storage instead. I believe the name CookieBanner is a classic name, based on old cookies policy, but now is broader, as recent privacy regulations must ask specifically for granular consents.

threequartersjohn:
> Note that cookies is just the way the consents are stored, we could be using local storage instead. I believe the name CookieBanner is a classic name, based on old cookies policy, but now is broader, as recent privacy regulations must ask specifically for granular consents. In a site where multiple granular consents must be given, this banner would have a checkbox for eacc consent.

Hmm, i understand. I'm ok with closing this PR if you don't think there is anything further to say on this.

satazor:
👍

PR #117, opened 2021-01-14
Use new syntax to set environment variables in CI

acostalima:
The workflow which deploys to Vercel is currently failing:

Screenshot 2021-01-14 at 13 09 08

The `set-env` command was [deprecated](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) a few months ago and recently disabled. This PR updates the workflow to use the [new syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files).

### DIFF
--- a/.github/workflows/deploy-vercel.yml
+++ b/.github/workflows/deploy-vercel.yml
@@ -27,14 +27,14 @@ jobs:
         env:
           SITE_URL: https://next-with-moxy.vercel.app
         run: |
-          echo "::set-env name=SITE_URL::$SITE_URL"
+          echo "SITE_URL=$SITE_URL" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


acostalima:
Now it's failing due to lack of permissions. 😕

satazor (approved):

satazor:
@acostalima I don't think the env variable is passing trough the vercel cli call, as the CI is failing saying we don't have access.

acostalima:
@satazor I've just echoed `SITE_URL` in the failing step and the variable is indeed defined. Even when inlining `https://next-with-moxy-pr-117.vercel.app`, it fails.

satazor:
Fixed it!

